PR #3, opened 2024-04-10
docs: clarify what `otk.target` means

mvo5:
I found the existing explaination a bit hard to follow, this is my attempt to make it a bit clearer.


achilleas-k (approved):
I can't speak to the accuracy with respect to the original intent (I might have misunderstood it), but I do find the extra paragraph clarifies it the way I understand it.

supakeen (approved):
That's how it was meant. Thank you.

PR #5, opened 2024-04-10
RFC: use curl dollar (`${}`) for var substituion

mvo5:
With curl dollar var substition some ambitious cases become slightly easier to see without knowning about the var content, e.g.:
```
$foo.bar
```
could mean "replace $foo" and add ".bar" if `$foo` is a string or "use member bar of map $foo". With
```
${foo}.bar
${foo.bar}
```
this would become easier to see without having context what type the var is.

### DIFF
--- a/doc/directives.md
+++ b/doc/directives.md
@@ -42,7 +42,7 @@ otk.argument:
   - architecture
 ```
 
-## Usage of `$`
+## Usage of `${}`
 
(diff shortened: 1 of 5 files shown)


supakeen:
Yea, this makes things much clearer. Can you fix up the conflicts?

supakeen (approved):

PR #7, opened 2024-04-10
RFC: use otk files instead of otk.arguments

mvo5:
The use-case given (and maybe I'm missing other use-cases here) for the otk.arguments is currently "version" and "architecture".

We could do this instead via a `fedora-39-amd64.yaml` file that just define those arguments. The upside of this is that we can easily see what combinations of version and architecture we support by looking at the top-level dir. It also simplifies the spec and we would not have to worry about validating arguments, i.e. how do we know that "-Dversion=41" is valid or not (if not valid it would probably mean down the line an include based on the version cannot be found and that presents challenges for a nice error message).

It would also allow us to have symlinks when things do not change, e.g.
```
rhel-9.3.yaml -> rhel-9.4.yaml
```

But maybe I'm missing another use case for this? Even then I think the `fedora/39-amd64.yaml` has some merrits and is worth considering.

### DIFF
--- a/doc/directives.md
+++ b/doc/directives.md
@@ -29,19 +29,6 @@ otk.define:
   boot_mode: uefi
 ```
 
-## otk.argument
-
-Define arguments that **MUST** be passed on the command line to `otk` with
-`otk compile -A`.
-
-Expects a `seq` for its value.
-
-```yaml
-otk.argument:
-  - version
-  - architecture
-```
-
 ## Usage of `${}`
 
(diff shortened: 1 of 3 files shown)


supakeen:
We can drop the `otk.argument` directive until there's a clear usecase for it. Can you update this PR?

achilleas-k (approved):
I really like this simplification.  It reduces the scope without limiting functionality.

PR #9, opened 2024-04-10
RFC: simplify customizations in the tree.yaml

mvo5:
For a lot of the OS customizations we really only change a single option. It would be nice to express this in a shorter way and I tried to capture it in this commit. Essentially the customization just moves a level deeper which AIUI is something that is already allowed with the current way customizations are specified.


supakeen:
Yea, this is correct; directives can appear anywhere in the tree unless otherwise specified (which I think we don't do very explicitly anywhere, I filed: #12 for this).

supakeen (approved):

PR #15, opened 2024-04-15
RFC: drop the short-form of variable

supakeen:
The short form of variable is not necessary yet.

### DIFF
--- a/doc/directives.md
+++ b/doc/directives.md
@@ -48,10 +48,8 @@ Use a previously defined variable. String values can be used inside other
 string values, non-string values *must* stand on their own.
 
 ```yaml
-otk.define:
-  variable: "foo"
-
 otk.include: ${variable}
+otk.include: path/${variable}.yml
 ```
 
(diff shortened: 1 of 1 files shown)


doc/directives.md
```diff
@@ -48,10 +48,8 @@ Use a previously defined variable. String values can be used inside other
 string values, non-string values *must* stand on their own.
 
 ```yaml
-otk.define:
-  variable: "foo"
-
 otk.include: ${variable}
```
mvo5, line 51: Sorry, I must miss something here, the example looks now incomplete to me, it uses `variable` but that is not defined, where/how is it defined? The PR descriptions says `drop the short-form of variable` but I don't see anything in but the example changing, could you elaborate bit more what this change is about (sorry, maybe I just need another cup of tea :)

supakeen: No, *I* needed a cup of coffee this morning; I read it as `otk.variable:` :)

mvo5: Thank you :)

PR #16, opened 2024-04-15
RFC: remove argument directive

supakeen:
We don't have a good usecase for arguments (yet).


supakeen:
Was already implemented in #7.

PR #22, opened 2024-04-19
doc: clarify directive placement

### LINKED ISSUE #12: Clearly document where directives are allowed
Directives are allowed anywhere in the tree unless otherwise specified. We don't write this down very clearly but we should mention this for each directive.

supakeen:
Closes #12.

### DIFF
--- a/doc/directives.md
+++ b/doc/directives.md
@@ -1,8 +1,8 @@
 # Directives
 
 `otk` has various directives that can be used in an omnifest. Generally these
-directives can appear anywhere in the tree and they are replaced with other
-trees or values.
+directives can appear anywhere in the tree unless otherwise specified and they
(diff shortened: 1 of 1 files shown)


doc/directives.md
```diff
@@ -1,8 +1,8 @@
 # Directives
 
 `otk` has various directives that can be used in an omnifest. Generally these
-directives can appear anywhere in the tree and they are replaced with other
-trees or values.
+directives can appear anywhere in the tree unless otherwise specified and they
```
schuellerf, line 4:
```suggestion
directives can appear anywhere in the tree unless otherwise specified (see below) and they
```
those exceptions would be "specified" in this documentation, right?

schuellerf (changes requested):

schuellerf (approved):

PR #23, opened 2024-04-19
pre-commit: introduce pre-commit hooks

schuellerf:
those check for eol, yaml syntax and similar

adding pytest is probably too slow and annoying... let's see

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ venv
 __pycache__
 .coverage
 htmlcov
+dictionary.dic
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -12,3 +12,8 @@ repos:
     rev: v1.35.1
     hooks:
       - id: yamllint
+  - repo: https://github.com/schuellerf/pre-commit-pyspelling
+    rev: 0.1.0
+    hooks:
+      - id: pyspelling
+        args: ["--config", ".spellcheck.yml"]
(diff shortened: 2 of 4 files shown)


supakeen:
@schuellerf Could we re-use the current `make` commands (`lint`, `test`, `type`) at least partially (I know `type` is failing atm)?

schuellerf:
pre-commit is different as it hooks into "git commit"
That's somehow different and I'd rather target to cover the basic github actions here to avoid roundtrips in PR-pushes

mvo5:
> @schuellerf Could we re-use the current `make` commands (`lint`, `test`, `type`) at least partially (I know `type` is failing atm)?

I agree here, it seems nice to have a single "make" or "make check" to run all checks (or is pre-commit maybe super common and I'm just not up-to-date?).

schuellerf:
It would just be an easy option to include it into the git workflow and not to forget the small things before committing.
still "make check" could run "pre-commit run --all-files" _and_ include the more extensive pytests

supakeen:
I'd be happy to merge this; we can iterate on it later as it's purely local developer experience. Can you add `pre-commit` to the dev dependency extras? :)

schuellerf:
> I'd be happy to merge this; we can iterate on it later as it's purely local developer experience. Can you add `pre-commit` to the dev dependency extras? :)

… totally makes sense - sorry :-D

supakeen (approved):

PR #25, opened 2024-04-19
test: implement simple sugar test

schuellerf:
I implemented some test, just to understand the project a little bit.

the expected_error probably will (should) improve? - but that can be done iteratively I guess

... and is this on purpose that resolve() changes the given variable in place?

### DIFF
--- a/test/test_sugar.py
+++ b/test/test_sugar.py
@@ -1,0 +1,61 @@
+import pytest
+
+from otk.parse.document import Omnifest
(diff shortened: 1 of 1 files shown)


supakeen:
Currently it does everything in-place though we will likely switch to copying the tree. If you use the return value of resolve in your test case it'll keep working after the change!

supakeen (approved):

PR #30, opened 2024-04-23
transform: copy the tree instead of in-place

supakeen:
This copies the tree for each transform giving the lower lying transforms a new tree every time.


schuellerf (approved):
I think it's good and easier to use - we'll tune the performance later if needed 😁

PR #33, opened 2024-04-24
[RFC] examples: split out "fedora/minimal-common.yaml" and define two archs

mvo5:
Tweak the example to be slightly more realistic and have two arches for fedora and a common yaml. This will probably see more refactor when an additonal fedora that is not minimal is added.

Tiny starting point, if this looks okay I will play a bit more with how fedora maybe in two arches and two flaviors could look like :)

### DIFF
--- a/example/fedora/minimal-40-aarch64.yaml
+++ b/example/fedora/minimal-40-aarch64.yaml
@@ -2,6 +2,6 @@ otk.version: 1
 
 otk.define:
   version: 40
-  architecture: x86_64
+  architecture: aarch64
 
 otk.include: "minimal-common.yaml"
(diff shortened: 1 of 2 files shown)


supakeen:
This is nice; you can also take some inspiration from the last big branch to be merged: https://github.com/supakeen/otk/tree/targets/example/fedora

mvo5:
> This is nice; you can also take some inspiration from the last big branch to be merged: https://github.com/supakeen/otk/tree/targets/example/fedora where I was trying out some similar organisation but also have a lot of hacky bits due to testing new features of that branch.

Cool, how do you want to do this, do you want to split this out from your tree and open separately or should we merge/keep this one and then rebase the other one on top of this?

supakeen (approved):
This one can be merged but you'll need to appease the linter.

supakeen (approved):

PR #52, opened 2024-05-07
examples: de-duplicate, add one more arch

mvo5:
Without at least two consumers of `minimal-common.yaml` the extraction
of the common definitions is a bit pointless :) So this commmits
adds one more consumer.

---

Recent commits adds content from "minimal-common.yaml" to the
"minimal-40-aarch64.yaml". This commit de-duplicates that again.

[I'm not sure I got this 100% correct, please double check that I really transferred the pieces in the right places]

### DIFF
--- a/example/fedora/minimal-40-aarch64.yaml
+++ b/example/fedora/minimal-40-aarch64.yaml
@@ -3,49 +3,5 @@ otk.version: 1
 otk.define:
   version: 40
   architecture: "aarch64"
-  packages:
-    # Repositories to fetch packages from
-    repositories:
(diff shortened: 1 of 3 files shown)


mvo5:
Actually the original https://github.com/osbuild/otk/pull/52 had two arches already (I was wondering) and one got removed somewhere along the way [0]. I understand the desire to move fast here but it feels we are a bit too fast and loose here :/

[0] https://github.com/osbuild/otk/commit/41df207f4c840a120e9b5ab311bfbcee43609c66 - no reason in the commit log why the removal happend.

supakeen (approved):

Issue #2, opened 2024-04-10
Please add a `otk.meta` example


mvo5:
I'm struggling a bit to understand the `otk.meta` directive. Right now it says:
```
Under the `otk.meta` namespace any data can be stored that other applications
need to use from an omnifest.

Expects a `map` for its value.

otk.meta.osbuild-composer:
  boot_mode: uefi
```
I would love an example what is done with the data, is it put into the generated manifest.json? is it just part of the .otk file for other tools to inspect?

Related to this to me it is not clear why it's it's own directive. I wonder if we could define the metadata just via the normal `otk.define`.
So
```yaml
otk.define:
 meta:
  osbuild-composer:
   export: image
```
i.e. I think I would love to have some example/explaination how this will be used.

supakeen:
This is going to be superceded by #85.

PR #53, opened 2024-05-07
RFC directive: remove `otk.meta` and use `target.*` instead

mvo5:
This is (very much) a RFC/brainstorm commit. Right now the use of `otk.meta` is a bit unclear, there is an open bug about this (c.f. https://github.com/osbuild/otk/issues/2).

But it seems to me that `meta` is really about writing json data from an omnifest for a specific tool (in this case "composer"). Which makes me think that we have support for this already with out `target` directive. So `otk -t osbuild-composer` would generate the needed meta information.

### DIFF
--- a/doc/03-omnifest/01-directive.md
+++ b/doc/03-omnifest/01-directive.md
@@ -173,27 +173,6 @@ otk.define:
         - ${b}
 ```
 
-## `otk.meta.<name>`
-
-Under the `otk.meta` namespace any data can be stored that other applications
-need to use from an omnifest.
-
-Expects a `map` for its value.
-
-```yaml
-otk.meta.osbuild-composer:
-  boot_mode: uefi
-  export: image
-  pipelines:
-    build:
-      - root
-    system:
-      - tree
-
-otk.meta.kiwi:
-  label: "A Label"
-```
-
 ## `otk.customization`
 
 Customizations are conditional blocks that receive separate input through
--- a/doc/07-integration.md
+++ b/doc/07-integration.md
@@ -6,7 +6,7 @@
 
 The osbuild-composer integration allows for [omnifests](./03-omnifest/index.md) to be consumed by images. This provides an API and jobqueue to facilitate building artifacts in Image Builder, Image Builder Frontend, and Cockpit Composer.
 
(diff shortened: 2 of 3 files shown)


supakeen:
I don't like this one, let's see what others think. 

Currently whatever lives under `otk.target.osbuild.<name>` is translated directly into the manifest and if we do this we'll have to start ignoring keys that are invalid in that schema. The target is also *not* `osbuild-composer`, we're generating an `osbuild` manifest.

The `otk.meta` keys are that; additional information for tools that aren't the target to allow them to make assumptions about what targets exist and what to do with them.

mvo5:
[..]
> Currently whatever lives under `otk.target.osbuild.<name>` is translated directly into the manifest and if we do this we'll have to start ignoring keys that are invalid in that schema. The target is also _not_ `osbuild-composer`, we're generating an `osbuild` manifest.

Maybe I'm misunderstanding but why would we start ignoring keys? The osbuild manifest would still be written via `otk.target.osbuild.*` and everything there needs to be valid. In additioon there is extra metadata that is *not* part of `osbuild` and *only* meaningful for `osbuild-composer` so for that extra meta-data the target is `osbuild-composer` as `osbuild` does not know about it or need it or can consume it. What am I missing or misrepresenting here?

> The `otk.meta` keys are that; additional information for tools that aren't the target to allow them to make assumptions about what targets exist and what to do with them.

But the `otk.meta` keys are only meaningful for a specific consumer, `otk.meta.*osbuild-composer*.boot_mode `is targeted to `osbuild-composer` - I would love to understand what is different between `target` and `meta` that we need them as distinct concepts (and if we add them as distinct concepts, *why* they are different). I.e. it seems to me by just commiting this change we could still generated the metadata for osbuild-composer by just passing `-t osbuild-composer` to the otk file? 

I may be wrong here of course, my knowledge about the metadata is from reading a bit in the composer source, I have not seen an actual file with an example :/

supakeen (commented):
I think we all agreed on doing this through targets instead; if you could fix the conflict I'll ack :)

supakeen:
I'll be updating the documentation in #264.

PR #56, opened 2024-05-07
02-external: move common text under depsolve-dnf{4,5}

mvo5:
The depsolve-dnf5 description was mixed up with the description below. This commit fixes this.

### DIFF
--- a/doc/03-omnifest/02-external.md
+++ b/doc/03-omnifest/02-external.md
@@ -4,19 +4,15 @@
 
 These directives are only allowed within a [`otk.target.osbuild.<name>`](./01-directive.md#otktargetconsumername).
 
-### `otk.external.osbuild.depsolve-dnf4`
+### `otk.external.osbuild.depsolve-dnf{4,5}`
 
 Solves a list of package specifications to RPMs and specifies them in the
 osbuild manifest as sources.
 
-### `otk.external.osbuild.depsolve-dnf5`
-
-Expects a `map` as its value.
+### `otk.external.osbuild.file-from-text`
 
 `osbuild` directives to write files. **If a stage exists for the type of file
 you want to write: use it.** See the [best practices](../04-best-practices.md).
-
-### `otk.external.osbuild.file-from-text`
 
 Write inline text to a file. Creates a source in the manifest and copies that
 source to the destination in the tree.


doc/03-omnifest/02-external.md
```diff
@@ -4,20 +4,16 @@
 
 These directives are only allowed within a [`otk.target.osbuild.<name>`](./01-directive.md#otktargetconsumername).
```
mvo5, line 5: If the directives are only allowed under a specific target, we could allow users to write this shother: `otk.external.osbuild.file-from-text`` -> `otk.external.file-from-text`` as we know that the context is `osbuild` from the  target. This would also ensure that users cannot use external commands that are not matching their target

doc/03-omnifest/02-external.md
```diff
@@ -4,20 +4,16 @@
...
-### `otk.external.osbuild.depsolve-dnf4`
+### `otk.external.osbuild.depsolve-dnf{4,5}`
 
 Solves a list of package specifications to RPMs and specifies them in the
 osbuild manifest as sources.
 
-### `otk.external.osbuild.depsolve-dnf5`
-
-Expects a `map` as its value.
+### `otk.external.osbuild.file-from-text`
```
mvo5, line 12: (nitpick) neither this nor the one below are actually used yet, I think would be ideal if all our directives have a use-case

doc/03-omnifest/02-external.md
supakeen: We could though I like the explicitness; I guess I'm about even on this so let's see what others feel like.

It'd *feel* weird to me if we had `otk.external.file-from-text` that does different things under different targets.

mvo5: I have no super strong opinion here but we should make sure that we do not allow using tools for targets that do not match the target target. 

I am not sure if e.g. different `otk.external.user` implementations depending on the target would be bad, I mean, they are clearly scoped under a
```
otk.target.osbuild:
 stages:
   - otk.external.users:
````
and in kiwi the same. But I don't really mind much, it just seems a bit redundant (both in terms of extra checks needed to avoid mixing targets and in extra typing for the users).

mvo5:
This is obsolete with the most recent discussions

PR #59, opened 2024-05-08
RFC: loops for customizations

mvo5:
When trying to port a slightly more elaborate customization to otk I noticed that many import ones (like users) are actually lists of items. Our synatx is currently not expressive enough for this.

There are (broadly) two ways to solve this:
1. run an external helper to deal with complex customizations
2. introduce a "loop" construct to the otk format

This is a stawman/draft for (2). I'm not sure how much I like it yet, I have some sympathy for (1) but I also think it's important to dicuss it and the loop itself does not look too bad.

### DIFF
--- a/doc/03-omnifest/01-directive.md
+++ b/doc/03-omnifest/01-directive.md
@@ -228,12 +228,12 @@ For this there is a `loop` construct:
 ```
 otk.customization.user:
   if-set:
-    # loop could imply "if-set"
+    # ensure /var/home on a bootc system
+    - type: org.osbuild.mkdir
+      options:
+        - path: /var/home/
     # loop loops over the customization sequence and yields "item" for each
     loop:
-      - type: org.osbuild.mkdir
-        options: 
-          - path: /var/home/${item.name}
       - type: org.osbuild.user
         options:
           ${item.name}:
--- a/doc/03-omnifest/01-directive.md
+++ b/doc/03-omnifest/01-directive.md
@@ -201,21 +201,49 @@ Customizations are conditional blocks that receive separate input through
 is passed data. If a customization is passed multiple times then the `defined`
 block is repeated multiple times, once for each input.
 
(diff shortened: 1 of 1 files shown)


supakeen:
My initial thoughts on this were to make the loops implicit. E.g. the `if-set` block will get done multiple times *if* the customization is 'passed' multiple times. Would we need a specific loop if we do that?

mvo5:
> My initial thoughts on this were to make the loops implicit. E.g. the `if-set` block will get done multiple times _if_ the customization is 'passed' multiple times. Would we need a specific loop if we do that?

Thank you! That is an interesting idea and will work in many (most? all?) cases. I tweaked the example to show one example in bib where it is useful to separate the two, i.e. in a bootc world if there is any user we need to create /var/home (only once) and then add a stage for each user. Of course here we could simply create /var/home all the time and it would not have ill effects but it's an example where having explicit control is beneficial.

In reality the bib case is actually more complicated because we need to either create /var/home or /var/roothome (or both) depending on the user(s). Which means to map this correctly we would need a real "if" potentially. But let's part this for a second but definitely something to further think about :)

mvo5:
Another way of course with jinja2 would be https://jinja.palletsprojects.com/en/3.1.x/templates/#for and this is not needed

mvo5:
Closing as this one was discussed recently in a video call and while it's similar to the proposal it's not the same so it's easier to just do a new PR:
```yaml
otk.op.loop:
  item: ${customization.files}
  foreach:
     - otk.external.osbuild.embed.file-from-text:
       text: ${item.text}
       path: ${item.path}
     - otk.op.if-set:
       what: ${item.mode}
       then:
         - type: org.osbuild.chmod
           options:
             path: ${item.path}
             mode: ${item.mode}
     - type: org.osbuild.chown
       options:
         path: ${item.path}
         user: ${item.user}
         group: ${item.group}
```

Issue #63, opened 2024-05-08
Map merging allows duplicate keys


supakeen:
Currently map merging will overwrite previous keys; this is in conflict with the specification which says this *should* be an error.

Let's make it an error.

Issue #62, opened 2024-05-08
Multiple directives clash with previous directives


supakeen:
When having multiple directives:

```yaml
otk.define:
  a: 1

otk.define:
  b: ${a}
```

Things go wrong. This is due to this being a mapping so duplicate keys at the same level are not allowed and have this behavior.

supakeen:
This *could* be solved if we were to do: #47 however it would annoy things like `yamllint` which will error on duplicate keys.

supakeen:
Note that doing this and processing duplicate keys would violate the principle of being valid YAML.

supakeen:
Also it *is* possible to work around this:

```yaml
otk.define:
  otk.op.join:
    - my: name
      my: other
    - otk.include: common.yaml
```

For example. Not the prettiest but it is explicit.

supakeen:
We're going to initially add a `tag` to includes in the form `otk.include.<tag>` this allows users to make the keys unique. This is probably not going to be final form.

PR #64, opened 2024-05-08
examples: fix up

supakeen:
The examples merged in #52 were broken (and I merged them knowing this) so now let's work to make the implementation support them.

Relevant bugs:

- #62
- #63


example/fedora/minimal-40-aarch64.yaml
```diff
@@ -1,7 +1,14 @@
...
 otk.define:
-  version: 40
-  architecture: "aarch64"
+  otk.op.join:
+    - version: 40
+      architecture: "aarch64"
+    - otk.include: "minimal-common.yaml"
 
-otk.include: "minimal-common.yaml"
+otk.target.osbuild:
```
mvo5, line 9: I wonder if we can do better here, not including the pipelines from something like "common.yaml" will mean we have quite a bit of duplication.

example/fedora/minimal-40-aarch64.yaml
```diff
@@ -1,7 +1,14 @@
 otk.version: 1
 
 otk.define:
-  version: 40
-  architecture: "aarch64"
+  otk.op.join:
```
mvo5, line 4: Is this change driven by the current implementation or is this the yaml we want? I'm asking because it does not feel like an improvement and I think at this stage of the project we should aim for the best possible yaml and then think about the implementation (and being able to have a toplevel with `otk.define` and an include that also `otk.defines` stuff without the need for an explicit `otk.op.join` seems nice.

mvo5 (commented):
I am not sure that the goal at this stage should be to have an implementation for our spec. I think the goal should be to create the best possible yaml via discussions and use-cases and then think about the implementation. I worry we make compromises too early otherwise (but happy to hear input from others here of course, every team does things differently).

example/fedora/minimal-40-aarch64.yaml
supakeen: This is driven by the current implementation and what is possible in it. And yes; the 'way it is' (in `main`) is nicer :)

example/fedora/minimal-40-aarch64.yaml
supakeen: We could resolve includes before omnifest validation.

supakeen:
Going to close this PR and resubmit due to shenanigans going on on GitHub/my local checkout.

PR #66, opened 2024-05-08
spec: Add a basic spec and Makefile targets

### LINKED ISSUE #44: Write a specfile
The project needs a specfile to be built as an RPM.

bcl:
This adds the srpm and rpm build makefile targets (borrowed from osbuild), and uses the pyproject rpm macros to support pyproject.toml The version is hard-coded into the spec so it will need to be changed at the same time as the version in pyproject.toml

Fixes #44

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,3 +15,33 @@ format:
 test:
 	@pytest
 	pre-commit run --all-files
+
+.PHONY: git-diff-check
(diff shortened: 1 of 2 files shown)


supakeen:
Awesome. Thank you.

supakeen (approved):

Issue #65, opened 2024-05-08
Add CI to compile examples


supakeen:
Let's add CI to compile the examples.

PR #80, opened 2024-05-14
CI: Add examples building for each PR / push

elkoniu:
OTK repo contains examples which can be used by users. This PR adds CI routine to be sure they are build each time without any errors.

Resolves: https://github.com/osbuild/otk/issues/65

### DIFF
--- a/.github/workflows/build-examples.yml
+++ b/.github/workflows/build-examples.yml
@@ -1,0 +1,27 @@
+name: Build examples
+
+on: [pull_request, push]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ["3.9", "3.10", "3.11", "3.12"]
+    steps:
+    - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


supakeen:
Feel free to fix up the `Self` import, I was unaware that this was such a recent addition to Python.

bcl:
Do we need to test 3.8? RHEL9 has python 3.9, and Fedora 39 and 40 are on python3.12 so we should probably test every version in-between 3.9 and 3.12

elkoniu:
> Feel free to fix up the `Self` import, I was unaware that this was such a recent addition to Python.

Can you elaborate on this a bit more? Which `self` import do you mean?

elkoniu:
> Do we need to test 3.8? RHEL9 has python 3.9, and Fedora 39 and 40 are on python3.12 so we should probably test every version in-between 3.9 and 3.12

I think we could skip 3.8 but it is up to @supakeen decision :)

supakeen:
@elkoniu 3.9 and up only.

supakeen:
This is now done through image references.

PR #82, opened 2024-05-14
examples: extract fedora-40-common.yaml with common values

mvo5:
Based on the most recent dicussion about includes with subkeys, let's de-duplicate the example.


supakeen:
Let's close this one; we'll likely redo all of the Fedora examples after the CentOS examples are done.

PR #83, opened 2024-05-15
hackathon: group coding session

supakeen:
Variety of small tidbits and examples as created during the Hackathon 00 group sessions.


example/centos/pipeline/os.yaml
```diff
@@ -0,0 +1,163 @@
...
+  - type: org.osbuild.sshd_config
+    options:
+      config:
+        PasswordAuthentication: ${sshd_password_authentication}
+  - otk.external.osbuild.partition-table.gen-fstab-stage:   # expands to "typ: org.osbuild.fstab\n..."
+  - type: org.osbuild.grub2
+    options:
+    root_fs_uuid: ${filesystem.root.uuid}
+    boot_fs_uuid: ${filesystem.boot.uuid}
+    kernel_opts: '{{ " ".join(kernel_opts_list) }}'
```
mvo5, line 147: This might need a "TODO: decide on final syntax"

mvo5 (dismissed):
Thanks for capturing this, this looks like exactly what we discuss.

mvo5:
Just for completeness, we will need to go over this again and de-duplicate fedora and centos pipelnes/yamls to see what can be shared and what has to be duplicated but that is obviously follows and we will need tests that compare any refactoring we do to the expected manifests (so for later, but mentioning it for posterity :)

supakeen:
Added more commits to this based on yesterdays work and discussion. Also did the straightforward change to use underscores in directive names.

PR #86, opened 2024-05-21
deps: Remove click dependency from `otk_osbuild`

elkoniu:
click module is not available in some distros.
This commit removes it usage from the otk_osbuild code.

### DIFF
--- a/src/otk_osbuild/command.py
+++ b/src/otk_osbuild/command.py
@@ -10,7 +10,7 @@ import base64
 import pathlib
 import subprocess
 
-import click
+import argparse
 
 
 def source_add(kind, data):
@@ -44,12 +44,6 @@ def source_add_curl(data, checksum, url):
         data["context"]["sources"]["org.osbuild.curl"]["items"][checksum] = {"url": url}
 
 
-@click.group()
-def root():
-    pass
-
-
-@root.command()
 def depsolve_dnf4():
     data = json.loads(sys.stdin.read())
     tree = data["tree"]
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -5,6 +5,7 @@ requires-python = ">= 3.9"
 
 dependencies = [
     "click >= 8.0",
+    "argparse >= 1.4",
```
supakeen, line 8: `argparse` is built-in, it can be dropped from the dependency list, see the [pypi project page](https://pypi.org/project/argparse/)

> As of Python >= 2.7 and >= 3.2, the argparse module is maintained within the Python standard library. For users who still need to support Python < 2.7 or < 3.2, it is also provided as a separate package, which tries to stay compatible with the module in the standard library, but also supports older Python versions.

src/otk_osbuild/command.py
```diff
@@ -191,3 +183,15 @@ def file_from_path():
...
     )
+
+
+def root():
+    parser = argparse.ArgumentParser()
+    subparsers = parser.add_subparsers()
+
+    (subparsers.add_parser('depsolve_dnf4')).set_defaults(func=depsolve_dnf4)
+    (subparsers.add_parser('file_from_text')).set_defaults(func=file_from_text)
+    (subparsers.add_parser('file_from_path')).set_defaults(func=file_from_path)
```
supakeen, line 194: If we're changing these to be underscores then the mapping also needs to be updated in `src/otk/external.py` at the bottom.

elkoniu: If the naming convention has changed? I keept the original function names. Shall I convert them to `_depsolve_dnf4` etc?

supakeen (approved):

Issue #88, opened 2024-05-27
Clean up the current depsolving proof of concept implementation


supakeen:
The current implementation of the `otk.osbuild.external.osbuild.depsolvex` is quite ... interesting. Let's clean it up from its poc state.

This requires #87 to be done.

PR #94, opened 2024-05-28
describe and implement the external protocol

supakeen:
Describes an initial external protocol.


doc/03-omnifest/02-external.md
```diff
@@ -1,22 +1,67 @@
...
-## osbuild
+External directives are directives that are implemented externally from `otk`.
+They are meant to be used to provide target-specific behavior and can be used
+to remove repetition and complexity.
+
+## Protocol
+
+Directives are small executables that receive JSON and are expected to output
+JSON again in a specific format. `otk` will look for these directives in
+`/usr/libexec/otk`. Externals are explicitly mapped to executables in the
```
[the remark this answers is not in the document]
supakeen: That's a good question. I don't know, what can we call it that would make sense across distributions?

doc/03-omnifest/02-external.md
```diff
@@ -1,22 +1,67 @@
...
+registry that is defined in `src/otk/external.py`.
+
+An external gets called with the following JSON:
+
+```json
+{
+  "tree": {
+    "otk.external.name": {
+      "child": [],
```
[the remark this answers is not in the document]
supakeen: They are indeed examples without special meaning. I'll add a YAML blurb!

doc/03-omnifest/02-external.md
```diff
@@ -1,22 +1,67 @@
...
+{
+  "tree": {
+    "what": {
+      "you": "want"
+    }
+  }
+}
+```
+
+*If the external returns `{}`, an empty object, `otk` will assume that there is
```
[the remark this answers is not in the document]
supakeen: It means the former. If it gets the latter then it will put a `{}` (empty object) into the output, I'll add an example of that.

doc/03-omnifest/02-external.md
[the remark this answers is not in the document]
supakeen: Dove into this a bit; from a quick test it seems Debian, Ubuntu, RHEL (UBI), Fedora, CentOS, Gentoo all have `/usr/libexec`. Alpine and Arch do not.

We can either do our own search path (`/usr/libexec` and then `/usr/lib`, perhaps with `/usr/local`-variants for each) or let eventual packages for Alpine or Arch add a patch.

doc/03-omnifest/02-external.md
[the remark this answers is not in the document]
supakeen: YAML blurb has been added.

doc/03-omnifest/02-external.md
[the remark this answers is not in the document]
supakeen: I've added YAML examples for this.

doc/03-omnifest/02-external.md
```diff
@@ -1,22 +1,67 @@
...
 
-## osbuild
+External directives are directives that are implemented externally from `otk`.
+They are meant to be used to provide target-specific behavior and can be used
+to remove repetition and complexity.
+
+## Protocol
+
+Directives are small executables that receive JSON and are expected to output
+JSON again in a specific format. `otk` will look for these directives in
```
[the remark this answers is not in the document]
supakeen: I've added a Paths section that includes this information.

doc/03-omnifest/02-external.md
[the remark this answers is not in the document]
supakeen: OK, I've updated the docs to search all relevant paths starting with an environment-defined one.

doc/03-omnifest/02-external.md
```diff
@@ -11,13 +11,22 @@ JSON again in a specific format. `otk` will look for these directives in
 `/usr/libexec/otk`. Externals are explicitly mapped to executables in the
 registry that is defined in `src/otk/external.py`.
 
-An external gets called with the following JSON:
+When we have an [omnifest](./index.md) that looks like this:
```
achilleas-k, line 12: Bit of a rephrase to avoid "you", "we", etc could be:
```
Given the following omnifest:
...

The external called `name` gets called with the following JSON as input:
...
```

doc/03-omnifest/02-external.md
```diff
@@ -1,22 +1,67 @@
...
+External directives are directives that are implemented externally from `otk`.
+They are meant to be used to provide target-specific behavior and can be used
+to remove repetition and complexity.
+
+## Protocol
+
+Directives are small executables that receive JSON and are expected to output
+JSON again in a specific format. `otk` will look for these directives in
+`/usr/libexec/otk`. Externals are explicitly mapped to executables in the
+registry that is defined in `src/otk/external.py`.
```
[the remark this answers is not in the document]
supakeen: I've removed this section and we can do lookups as described in the new Paths section.

doc/03-omnifest/02-external.md
```diff
@@ -1,22 +1,67 @@
 # External
 
-## osbuild
+External directives are directives that are implemented externally from `otk`.
+They are meant to be used to provide target-specific behavior and can be used
+to remove repetition and complexity.
```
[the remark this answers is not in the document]
supakeen: I've added this.

doc/03-omnifest/02-external.md
```diff
@@ -1,22 +1,67 @@
...
+      "options": "are here"
+    }
+  }
+}
+```
+
+And is expected to return the following JSON:
+```json
+{
+  "tree": {
```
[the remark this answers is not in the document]
supakeen: I've mentioned that additional keys are not allowed.

doc/03-omnifest/02-external.md
```diff
@@ -1,22 +1,117 @@
...
+    - 1
+  options: "are here"
+```
+
+The external gets called with the following JSON:
+
+```json
+{
+  "tree": {
+    "otk.external.name": {
```
mvo5, line 26: Curious, why do we include the name in the tree? In case an external is symlinked and would change "personality" based on the inputs? But even then it would know this from argv0, no?

doc/03-omnifest/02-external.md
```diff
@@ -1,22 +1,117 @@
...
+  otk.external.name:
+    options:
+```
+
+Into this YAML structure:
+
+```yaml
+dummy:
+```
+
```
achilleas-k, line 69: Perhaps a "hello world" style example would be nice here too, something that's fully working but doesn't do anything interesting.

I'm thinking something like:

## Example

The following example demonstrates how an external can be used to implement simple string concatenation.

Given the following script, called `concat`, in `/usr/local/libexec/otk/concat`
```bash
#!/usr/bin/env bash

output="$(jq -jr '.tree."otk.external.concat".parts[]' <<< "${1}")"
echo "{\"tree\":{\"output\":\"$output\"}}"
```

and the following directive:
```yaml
examplestring:
  otk.external.concat:
    parts:
      - list
      - of
      - strings
```

the script is called with the following input:
```json
{
  "tree": {
    "otk.external.concat": {
      "parts": [
        "list",
        "of",
        "strings"
      ]
    }
  }
}
```

which results in the following output:
```json
{
  "tree": {
    "string": "listofstrings"
  }
}
```

and the final omnifest will be:
```yaml
examplestring:
  listofstrings
```

achilleas-k (commented):
I think a

doc/03-omnifest/02-external.md
```diff
@@ -1,22 +1,117 @@
...
+Which will lead to the following YAML structure:
+
+```yaml
+tree:
+  what:
+    you: "want"
+```
+
+If the external returns `{}`, an empty object, `otk` will assume that there is
```
mvo5, line 54: What will happen if `{tree: {}}` is returned?

doc/03-omnifest/02-external.md
```diff
@@ -1,22 +1,117 @@
...
+it finds the first match:
+
+- A path defined by the `OTK_EXTERNAL_PATH` environment variable.
+- `/usr/local/libexec/otk`
+- `/usr/libexec/otk`
+- `/usr/local/lib/otk`
+- `/usr/lib/otk`
+
+The filename for an external executable is based on the external name. When the
+following directive is encountered: `otk.external.<name>.<argument>` then
```
mvo5, line 82: Do you have an example for `<argument>` ? I'm not sure I follow why we need this mapping. As a stawman, maybe something very simple. The search path is the same as above but contains "/external/" as the last component (so /usr/libexec/otk/external" and then just `osbuild_dnf4" ? Or are we limiting ourselfs here is do something very simple like this?

mvo5 (commented):
Thank you! I like it, still some ideas/questions/suggestions inline, hope they are useful :)

doc/03-omnifest/02-external.md
achilleas-k: Is this meant to be like how git (and others) work with subcommands?

achilleas-k: I think in the interest of keeping things simple for the initial implementation, we can avoid adding features that can be expressed easily with other features, especially if the alternative isn't too complicated.  A subcommand here can easily be expressed using an option for the external.  

```yaml
...
  otk.external.myscript.subcommand_1:
    option: "value"
```

is not that much cleaner than:
```yaml
...
  otk.external.myscript:
    command: "subcommand_1"
    options: "value"
```

I think the fact that the first form requires an extra section in the specification and docs is a sign that it's not necessary given how the alternative isn't really that complicated.

doc/03-omnifest/02-external.md
```diff
@@ -1,22 +1,117 @@
...
+`otk` will try to find an executable called `otk_external_<name>` in the previously
+mentioned search paths.
+
+It will call this executable with `<argument>` as its first argument. Here are
+some examples:
+
+- `otk.external.foo` -> `otk_external_foo`
+- `otk.external.foo.bar_baz` -> `otk_external_foo bar_baz`
+- `otk.external.foo.bar.baz` -> `otk_external_foo bar baz`
```
achilleas-k, line 91: Already I'm looking at this and thinking "is `bar baz` a single argument with a space or two args?"
If it's the former, how do I do two args?
If it's the latter, how do I do args with spaces?

Neither of these questions exist if the specification doesn't allow arguments in the `otk.external` directive's name and subcommands/arguments just become part of the input.

supakeen: Yep, I'm going to drop it, it's confusing.

supakeen: Dropped.

doc/03-omnifest/02-external.md
supakeen: Done.

doc/03-omnifest/02-external.md
supakeen: It's to make it a bit more explicit to show which part of the tree it is operating on.

doc/03-omnifest/02-external.md
```diff
@@ -1,22 +1,117 @@
...
+      "you": "want"
+    }
+  }
+}
+```
+
+Which will lead to the following YAML structure:
+
+```yaml
+tree:
```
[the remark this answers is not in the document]
supakeen: Bad name, I added a prefix to make it clear where the replacement happens.

doc/03-omnifest/02-external.md
```diff
@@ -1,22 +1,117 @@
...
+{
+  "tree": {
+    "what": {
+      "you": "want"
+    }
+  }
+}
+```
+
+Which will lead to the following YAML structure:
```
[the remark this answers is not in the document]
supakeen: Done.

doc/03-omnifest/02-external.md
```diff
@@ -1,22 +1,117 @@
...
+  "tree": {
+    "otk.external.name": {
+      "child": [1],
+      "options": "are here"
+    }
+  }
+}
+```
+
+And is expected to return the following JSON:
```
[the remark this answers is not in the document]
supakeen: Done.

doc/03-omnifest/02-external.md
supakeen: It'd put a `{}` (empty object) at that spot in the tree.

doc/03-omnifest/02-external.md
supakeen: This is now included in the PR.

doc/03-omnifest/02-external.md
```diff
@@ -67,6 +67,66 @@ Into this YAML structure:
...
+#!/usr/bin/env bash
+
+output="$(jq -jr '.tree."otk.external.concat".parts[]' <<< "${1}")"
+echo "{\"tree\":{\"output\":\"$output\"}}"
+```
+
+and the following directive:
+
+```yaml
+and the following directive:
```
achilleas-k, line 87: Duplicate line snuck into the yaml code block.

doc/03-omnifest/02-external.md
achilleas-k: Still there?

achilleas-k (changes requested):
LGTM generally.  Tiny correction for a duplicate line.
Also, the subcommand/argument stuff and rephrase from my previous comments seem unchanged?

doc/03-omnifest/02-external.md
supakeen: Removed.

doc/03-omnifest/02-external.md
supakeen: Dropped, I think I only dropped it form the code implementation initially.

achilleas-k (approved):
LGTM

doc/03-omnifest/02-external.md
```diff
@@ -1,22 +1,117 @@
...
+```json
+{
+  "tree": {
+    "otk.external.name": {
+      "child": [1],
+      "options": "are here"
+    }
+  }
+}
+```
```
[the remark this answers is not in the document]
mvo5: Super nitpick - the initial part of the suggestion to change ` When we have an [omnifest](./index.md) that looks like this:` -> `Given the following [omnifest](https://github.com/osbuild/otk/pull/94/index.md):` seems to be gotten lost (which is not a big deal but I wanted to mention it :)

doc/03-omnifest/02-external.md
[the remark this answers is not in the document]
mvo5: Similar to above, this says "Done" but AFAICT the text is unchanged, is it "done" in an indirect way?

src/otk/external.py
```diff
@@ -1,85 +1,73 @@
...
-
-    required_context, executable = registry[directive]
-
-    if not isinstance(context, required_context):
-        log.fatal("external %r with wrong context %r", directive, context)
-
-        # TODO exception
-        sys.exit(1)
-        return
+    exe, args = convert(directive)
```
mvo5, line 20: AFAICT we do not use "args" currently, maybe worth removing them until we need them? then this could become something like `exe = exe_from_directive(directive)`  or `exe = exe_from(directive)`

src/otk/external.py
```diff
@@ -1,85 +1,73 @@
...
-    context.from_external(data["context"])
 
-    # TODO restore context
-    return data["tree"]
+def convert(directive):
+    exe, *args = directive.removeprefix(PREFIX_EXTERNAL).split(".")
+    return (f"otk_external_{exe}", args)
+
+
+def search(exe):
```
mvo5, line 54: (nitpick) `search()` is a very generic name, maybe `path_for(exe)` or `path_for_external(exe)` or something?

src/otk/external.py
```diff
@@ -1,85 +1,73 @@
...
+    paths = [
+        "/usr/local/libexec/otk",
+        "/usr/libexec/otk",
+        "/usr/local/lib/otk",
+        "/usr/lib/otk",
+    ]
+
+    env = os.getenv("OTK_EXTERNAL_PATH", None)
```
mvo5, line 62: I wonder if `OTK_EXTERNAL_PATH` should support `:`? But I guess we can always add it later if needed :)

test/test_external.py
```diff
@@ -0,0 +1,13 @@
+import pytest
+
+from otk.external import convert
+
+
+@pytest.mark.parametrize("text,exe,args", [
```
mvo5, line 6: Is the json handling tested elsewhere or should there be a followup with some more testing (afaict `external.py:call()` is currently untested(?)).

mvo5 (commented):
Thanks for the changes, the README looks very nice now! Fine to merge IMHO. 

I have some questions in the implementation though, maybe I'm misunderstanding something but it seems like the parsing of external names to an executable and args for the executable was removed from the spec but the code seems to be still doing it? Is that because the otk_external_osbuild is easier to implement this way? i.e. instead of looking at "os.argv[0]" how it got called it would be called via the argument magic from "convert()"? If so we should reconsider if we want it back in the spec or if we want to change the code to look at argv[0] but it seems right now there is a bit of a mismatch between spec and code.

doc/03-omnifest/02-external.md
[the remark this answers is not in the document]
supakeen: Some commit seems to have gone missing, really done now :)

doc/03-omnifest/02-external.md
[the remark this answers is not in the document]
supakeen: Some commit seems to have gone missing, really done now :)

src/otk/external.py
```diff
@@ -1,85 +1,73 @@
...
+
+    if env is not None:
+        paths = [env] + paths
+
+    for path in paths:
+        path = pathlib.Path(path) / exe
 
+        if path.exists():
```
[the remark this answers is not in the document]
supakeen: Good idea, done.

src/otk/external.py
supakeen: Done.

test/test_external.py
supakeen: Some more testing should eventually happen :)

src/otk/external.py
```diff
@@ -1,85 +1,73 @@
...
+            args,
             directive,
             process.stdout,
             process.stderr,
         )
         sys.exit(1)
         return
 
-    # Otherwise everything is good and we restore
-    data = json.loads(process.stdout)
```
mvo5, line 60: Hm, unless I am missing something this seems to be still needed or else we never load the new data from the external(?). It feels like we want to have a bit more test coverage here (see my other comment that has a small starting point commit).

mvo5 (changes requested):
Thanks for the update! I did a more careful review of the python code this time and noticed what appears to be a small oversight in the way the data from the external helper is loaded in call (hence the change request). My other comments are just ideas or suggestions :)

src/otk/external.py
supakeen: Done.

src/otk/external.py
```diff
@@ -1,85 +1,73 @@
...
 from typing import Any
 
-from .context import Context, OSBuildContext
+from .context import Context
+from .constant import PREFIX_EXTERNAL
 
 log = logging.getLogger(__name__)
 
 
 def call(context: Context, directive: str, tree: Any) -> Any:
```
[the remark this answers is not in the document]
supakeen: It can be removed.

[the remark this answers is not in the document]
supakeen: Done.

src/otk/external.py
```diff
@@ -1,85 +1,73 @@
...
-    "otk.external.osbuild.file_from_text": (
-        OSBuildContext,
-        ["otk-osbuild", "file-from-text"],
-    ),
-    "otk.external.osbuild.file_from_path": (
-        OSBuildContext,
-        ["otk-osbuild", "file-from-path"],
-    ),
-}
+    raise RuntimeError(f"could not find {exe!r} in any search path {paths!r}")
```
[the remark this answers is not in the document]
supakeen: The path from the environment is included in `paths` :)

src/otk/external.py
supakeen: This is still needed, I'll add the test you wrote then fix it!

src/otk/external.py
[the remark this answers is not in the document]
mvo5: Sorry, my bad, I misread the code - all is fine then, thank you!

doc/03-omnifest/02-external.md
mvo5: Fwiw, the spec and the (current) implementation diverge here, when adding the following test:
```python
def test_integration_happy(tmp_path):
    os.environ["OTK_EXTERNAL_PATH"] = os.fspath(tmp_path)
    fake_external_path = tmp_path / "otk_external_test"
    fake_external_path.write_text(textwrap.dedent("""\
    #!/bin/sh
    cat - > "$0".stdin
    echo '{"tree": {"some": "result"}}'
    """))
    os.chmod(fake_external_path, 0o755)

    fake_directive = "otk.external.test"
    fake_tree = {
        "foo": "bar",
        "sub": {
            "baz": "foobar",
        },
    }

    res = otk.external.call(fake_directive, fake_tree)
    assert res == {"some": "result"}

    inp = fake_external_path.with_suffix(".stdin").read_text()
    assert json.loads(inp) == {"tree": fake_tree}
```
(note the last two lines that are new compared to the patch I attached to this originally).

Then it shows that the `otk.external.name` is (currently) not part of the inputs. Not a big deal but I wanted to mention it so that we are aware :)

mvo5 (commented):
Thanks for the updates! I went over it again and also updated the small integration test for the externals, I added checks in inputs too (sorry, forgot to do that yesterday). It's attached here: 
[0001-external-write-integration-test-for-the-happy-case.patch.txt](https://github.com/user-attachments/files/15576034/0001-external-write-integration-test-for-the-happy-case.patch.txt)

achilleas-k:
So this PR is getting a bit hard to re-re-review after all the comments and follow-up commits.  Would it be difficult to have the commit history cleaned up a bit?  For example, 062ed9015eff3a984bb4ac70d8a8c190c33eadf9 could be squashed into 8e172c9e4a084c3714f1cc49f8098fb8b3227ad8 (the commit that drops the usage of the `OSBuildContext`).  Smaller commits like the 4 that follow the first can be squashed into the first one that's the bulk of the doc writing.  Stuff like that.

doc/03-omnifest/01-directive.md
```diff
@@ -200,7 +200,7 @@ otk.meta.kiwi:
 
 External directives. Directives starting with `otk.external` are redirected
 to `/usr/libexec/otk/`-binaries. For example the directive
-`otk.external.osbuild.depsolve_dnf4` will execute `otk-osbuild depsolve_dnf4`
+`otk.external.osbuild_depsolve_dnf4` will execute `otk_external_osbuild_depsolve_dnf4`
```
mvo5, line 203: Slightly silly question - there is nothing here that forces us to use names like `osbuild_` we could keep the previous `osbuild.` as long as we name the executable files accordingly?

src/otk/external.py
```diff
@@ -1,83 +1,73 @@
...
-    required_context, executable = registry[directive]
-
-    if not isinstance(context, required_context):
-        log.fatal("external %r with wrong context %r", directive, context)
-
-        # TODO exception
-        sys.exit(1)
-        return
+def call(directive: str, tree: Any) -> Any:
+    exe = convert(directive)
```
mvo5, line 19: (super nitpick) I wonder if a slightly more specific name than `convert` would be easier to read here, maybe something like `exe = exe_from_directive(directive)` or `exe = exe_from(directive)`? (fwiw, this was marked as "done" in https://github.com/osbuild/otk/pull/94#discussion_r1625847169 but maybe got lost in some force push?)

src/otk/external.py
```diff
@@ -1,85 +1,73 @@
...
-    # TODO restore context
-    return data["tree"]
+def convert(directive):
+    exe, *args = directive.removeprefix(PREFIX_EXTERNAL).split(".")
+    return (f"otk_external_{exe}", args)
+
+
+def path_for(exe):
+    paths = [
+        "/usr/local/libexec/otk",
```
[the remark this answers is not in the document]
mvo5: Any opinion here? Happy to do that in a followup but would love to avoid it getting lost :)

src/otk_external_osbuild/command.py
```diff
@@ -46,7 +46,7 @@ def source_add_curl(data, checksum, url):
 
 def depsolve_dnf4():
     data = json.loads(sys.stdin.read())
-    tree = data["tree"]
+    tree = data["tree"]["otk.external.osbuild_depsolve_dnf4"]
```
mvo5, line 49: Dol we need a similar change for the two other externals we have (file_from_text and file_from_path)?

mvo5 (commented):
Thank you for the history cleanup! This looks pretty ready now, I have a few inline questions but those are no blockers so I'm fine approving if we want to do the open questions in a followup (happy to helper here of course :)

doc/03-omnifest/01-directive.md
supakeen: We could.

src/otk_external_osbuild/command.py
supakeen: Yes we do. Will be in followup under #88.

mvo5 (approved):
Thank you!

PR #98, opened 2024-05-28
ci: tag main builds with their git commit SHA

ondrejbudai:
This is useful when you want pin the image. Also, it's very helpful
to easily see which commit was used to build container image.


supakeen (approved):
Looks good, thank you.

PR #99, opened 2024-05-28
directive: define fix

supakeen:
`otk.define` is incorrectly handling overwriting values by not overwriting them at all. Let's fix it and add a test.

### DIFF
--- a/test/test_directive_define.py
+++ b/test/test_directive_define.py
@@ -1,0 +1,38 @@
+import pytest
+
+from otk.context import CommonContext
+from otk.error import TransformDirectiveTypeError
+from otk.directive import define
+
+
+def test_define():
+    ctx = CommonContext()
+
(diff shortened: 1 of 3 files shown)


test/test_directive_define.py
```diff
@@ -0,0 +1,38 @@
...
+
+from otk.context import CommonContext
+from otk.error import TransformDirectiveTypeError
+from otk.directive import define
+
+
+def test_define():
+    ctx = CommonContext()
+
+    define(ctx, {"a": "b", "c": 1})
```
bcl, line 11: This is really minor, but I find using 'b' for the value of 'a' makes it harder to read. If it were a word I wouldn't need to make sure that 'b' wasn't supposed to be a key.

Also, would it be worthwhile to have a key that doesn't get changed?

supakeen:
Going to merge first then update the tests so I can get a few more example manifests working :)

Issue #101, opened 2024-05-29
Specification for read-only (or const) defines


achilleas-k:
While discussing the partition table generator with @mvo5, we considered the idea of having a way to signal that certain defines shouldn't be modified after they're set (even producing an error if they are).  This is especially important for large, complicated data structures generated by externals where there are interdependencies between values (like the partition table).

Consider the following scenario:
- The partition table external generates a fully-described PT structure (with UUIDs, starts and offets, etc) and stores it in `${filesystems}`
- A truncate stage is generated as:
```yaml
otk.target.osbuild:
pipeline:
 disk:
  - type: org.osbuild.truncate
    options:
      filename: "whatever.disk"
      size: ${filesystems.partition_table.size}
```
- The user decides to change the size of the partition table but due to a poor understanding of the mechanics of externals and defines, does the following:
```yaml
otk.define.defaults:
  filesystems:
    otk.external.gen_partition_table:
      ...
      size: "10 GiB"
      ...
  
  ...
  # potentially in a different file
  filesystems.partition_table.size: "20 GiB"

otk.target.osbuild:
pipeline:
 disk:
  - type: org.osbuild.truncate
    options:
      filename: whatever.disk
      size: ${filesystems.partition_table.size}
```

If this is allowed, it'll throw off the the partition table consistency and partition positions and alignments.  Changing other values could also make the whole thing unbuildable (which might be better or worse depending on POV).

It would be nice to have a way to _protect_ values inside data structures generated by externals in some way, perhaps with a keyword convention (`const`?).  So in the example above, the generator adds the `const` (or whatever) keyword to the data structure it produces, making the  property `${fileystems.const...}` and anything under `const` can't be modified after it's set the first time and doing:
```yaml
otk.define.stuff:
  filesystems.const.partition_table.size: "5 KiB"
```
produces an error.

mvo5:
Fwiw, I think we don't need to enforce this programmatically initially, having a strong hint to the user that some defines should not be modified will probably be already very helpful (so picking a good name "const", "protected", "readonly" ....) is helpful :)

supakeen:
Is the current implementation of warning on overwrite enough (it will probably need to deal with nesting)?

supakeen:
(ignore the cruft):

```
(venv) user@muja otk € otk -w duplicate-definition compile test.yaml
WARNING:otk.context:redefinition of 'a', previous values was 1 and new value is 2
...
```

mvo5:
[...]

PR #102, opened 2024-05-29
otk: don't use click

supakeen:
Start removing click from the main otk command.


supakeen:
This PR also drops the `-w/--warn` option as we're still figuring out how to deal with that in #101.

example.yaml
```diff
@@ -1,9 +0,0 @@
-otk.version: "1"
-
-otk.define.tag:
-  var: 1
-
-otk.target.osbuild.example:
-  pipelines:
-    - otk.include.tag: include.yaml
-    - otk.include.bar: include.yaml
```
achilleas-k, line 9: Can we have the removals of these examples as a separate commit?  They're not related to the arg parser are they?

src/otk/command.py
```diff
@@ -19,144 +15,182 @@
...
-    and translate them into the native formats for image build tooling."""
+def root():
+    argv = sys.argv[1:]
+
+    parser = parser_create()
 
-    ctx.ensure_object(dict)
+    if not len(argv):
+        parser.print_help()
+        parser.exit()
```
achilleas-k, line 25: Is this necessary?  argparse supports making an argument or subcommand a requirement which will do exactly this (print help and exit) if none are given.

src/otk/command.py
```diff
@@ -19,144 +15,182 @@
...
-    ctx: click.Context,
-    input: str,
-    output: Optional[str],
-    external: bool,
-    requested_target: str,
-) -> None:
-    """Compile a given omnifest into its targets."""
+    if arguments.command is None:
+        parser.print_help()
+        parser.exit()
```
achilleas-k, line 49: Again, I think these can (and should) be handled using argparse functionality.

src/otk/command.py
```diff
@@ -19,144 +15,182 @@
...
 
-    log.info("Compiling the input file %r to %r", input, output or "STDOUT")
+    # a subcommand is *required*, argparse has already handled non-existing
+    # subcommands
+    if arguments.command is None:
+        parser.print_help()
+        parser.exit()
```
achilleas-k, line 55: Same here.

achilleas-k (changes requested):

example.yaml
supakeen: Correct, they shouldn't have been in the project at all.

src/otk/command.py
supakeen: The default error provided by argparse when setting subcommands to required is more unclear to me than showing the full help:

```
(venv) user@nezhit otk € otk
usage: otk [-h] [-j] [-v] [-i IDENTIFIER] [-w WARN] {compile} ...
otk: error: the following arguments are required: command
```

I'd be fine with doing that if you feel that error is clear enough.

achilleas-k: Testing again, I guess we don't get the full help when things go wrong.  Not sure how much we want that vs what argparse produces (a list of valid values and a short usage line), but I think it'd be cleaner and easier to maintain if we relied on argparse functionality vs checking ourselves.

supakeen: Updated this to use built-in subcommand requirements.

src/otk/command.py
```diff
@@ -19,144 +15,169 @@
...
+        "/dev/stdin" if arguments.input is None else arguments.input
+    )
+    dst = pathlib.Path(
+        "/dev/stdout" if arguments.output is None else arguments.output
+    )
 
-    log.info("Compiling the input file %r to %r", input, output or "STDOUT")
+    if not src.exists():
+        log.fatal("INPUT path %r does not exist", str(src))
+        return 1
```
achilleas-k, line 60: Function's type annotation should say it returns an `int` now.

src/otk/command.py
```diff
@@ -19,144 +15,169 @@
...
+    # some arguments need to be handled before continuing with
+    # subcommands
+    if arguments.identifier and not arguments.json:
+        parser.print_help()
+        parser.exit()
+
+    if arguments.command == "compile":
+        return compile(parser, arguments)
+    else:
+        raise RuntimeError("Unknown subcommand")
```
achilleas-k, line 45: The else is unnecessary after the return, but nbd.

achilleas-k (commented):
Thanks!  LGTM except for a tiny comment and a type annotation.

Much cleaner now I think :)

src/otk/command.py
supakeen: Added a commit to fix the signature.

achilleas-k (approved):

PR #105, opened 2024-06-04
otk/command: implement "validate" command

### LINKED ISSUE #79: Add an `otk validate` subcommand
Let's add an `otk validate` subcommand which validates a given omnifest but produces no output. The exit code should be `0` for a valid omnifest and `1` for a non-valid omnifest.

schuellerf:
I also fixed the behavior of "output" as it was expected to be a path - which is wrong I guess
fixes #79

### DIFF
--- a/src/otk/command.py
+++ b/src/otk/command.py
@@ -21,7 +21,7 @@ def root() -> int:
     parser = parser_create()
     arguments = parser.parse_args(argv)
 
-    # turn on logging as *soon* as possible so it can be used early
+    # turn on logging as *soon* as possible, so it can be used early
     logging.basicConfig(
         level=logging.WARNING - (10 * arguments.verbose),
         handlers=[
@@ -40,26 +40,26 @@ def root() -> int:
         parser.exit()
 
     if arguments.command == "compile":
-        return compile(parser, arguments)
+        return compile(arguments)
+    elif arguments.command == "validate":
+        return validate(arguments)
 
     raise RuntimeError("Unknown subcommand")
 
 
-def compile(
-    parser: argparse.ArgumentParser, arguments: argparse.Namespace
-) -> int:
+def _process(arguments: argparse.Namespace, dry_run: bool) -> int:
     src = pathlib.Path(
         "/dev/stdin" if arguments.input is None else arguments.input
     )
     dst = pathlib.Path(
         "/dev/stdout" if arguments.output is None else arguments.output
-    )
+    ) if not dry_run else None
 
     if not src.exists():
         log.fatal("INPUT path %r does not exist", str(src))
         return 1
 
-    if not dst.exists():
+    if not dry_run and not dst.parent.exists():
         log.fatal("OUTPUT path %r does not exist", str(dst))
         return 1
 
@@ -107,7 +107,7 @@ def compile(
         kind, name = target_requested.split(".")
     except ValueError:
         # TODO handle earlier
-        log.fatal("malformatted target name %r", target_requested)
+        log.fatal("malformed target name %r. We need a format of '<TARGET_KIND>.<TARGET_NAME>'.", target_requested)
         return 1
 
     # re-resolve the specific target with the specific context and target if
@@ -116,11 +116,20 @@ def compile(
     tree = resolve(spec, tree[f"{PREFIX_TARGET}{kind}.{name}"])
 
     # and then output by writing to the output
-    dst.write_text(
-        target_registry.get(kind, CommonTarget)().as_string(spec, tree)
-    )
+    if not dry_run:
+        dst.write_text(
+            target_registry.get(kind, CommonTarget)().as_string(spec, tree)
+        )
 
     return 0
+
+
+def compile(arguments: argparse.Namespace) -> int:
+    return _process(arguments, dry_run=False)
+
+
+def validate(arguments: argparse.Namespace) -> int:
+    return _process(arguments, dry_run=True)
 
 
 def parser_create() -> argparse.Namespace:
@@ -178,6 +187,21 @@ def parser_create() -> argparse.Namespace:
         "--target",
         default=None,
         help="Target to output, required if more than one target exists in an omnifest.",
+    )
+
+    parser_validate = subparsers.add_parser("validate", help="Validate an omnifest.")
+    parser_validate.add_argument(
+        "input",
+        metavar="INPUT",
+        nargs="?",
+        default=None,
+        help="Omnifest to validate to or none for STDIN.",
+    )
+    parser_validate.add_argument(
+        "-t",
+        "--target",
+        default=None,
+        help="Target to validate, required if more than one target exists in an omnifest.",
     )
 
     return parser
--- a/test/test_command.py
+++ b/test/test_command.py
@@ -1,65 +1,68 @@
+import pytest
+
 from otk.command import parser_create
 
 
-def test_parse_no_command():
+def test_parse_no_command(capsys):
     p = parser_create()
+    with pytest.raises(SystemExit) as sys_exit:
+        p.parse_args([])
+    assert sys_exit.type == SystemExit
+    assert sys_exit.value.code == 2
 
-    r = p.parse_args([])
-
-    assert r.command is None
-    assert r.verbose == 0
-    assert not r.json
-    assert r.identifier is None
-
-    r = p.parse_args(["-j"])
-    assert r.command is None
-    assert r.json
-    assert r.verbose == 0
-    assert r.identifier is None
-
-    r = p.parse_args(["-j", "-v"])
-    assert r.command is None
-    assert r.json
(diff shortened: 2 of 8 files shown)


schuellerf:
as I ran into the linter in github - I also integrated ruff into pre-commit now.

please comment on those:
 - should this be a separate PR?
 - should we keep the "--fix" in pre-commit for `ruff`?

supakeen:
> as I ran into the linter in github - I also integrated ruff into pre-commit now.
> 
> please comment on those:
> 
>     * should this be a separate PR?
> 
>     * should we keep the "--fix" in pre-commit for `ruff`?

I'm fine with having this in the same PR, let's not overcomplicate since I agree with both additions.

There's just an architectural remark I have. I don't quite like `validate` calling the `compile` subcommand and I'd prefer having them as separate commands, we can move the common code to a helper function.

supakeen (changes requested):
I already wrote this in a normal comment; I'd like to keep the `compile` and `validate` commands separate and move their function bodies to a common helper.

mvo5:
Fwiw, I would be in favor of splitting purely whitespace changes (like the last (two?) commit(s) that just does a bunch of ruff formatting) into a separate PR. I think it's distracting otherwise for the reviewer (and for people looking at the PR history)

test/test_command.py
```diff
@@ -1,65 +1,46 @@
...
+    with pytest.raises(SystemExit) as sys_exit:
+        p.parse_args([])
+    assert sys_exit.type == SystemExit
+    assert sys_exit.value.code == 2
+
+    captured_stderr = capsys.readouterr().err
+    assert "the following arguments are required: command" in captured_stderr
+
+
+@pytest.mark.parametrize("command,results,sys_exit_code,sys_exit_message", [
```
mvo5, line 17: I like this!

test/test_command.py
```diff
@@ -1,65 +1,46 @@
...
-    r = p.parse_args(["compile", "-o", "file.yaml"])
-    assert r.command == "compile"
-    assert r.input is None
-    assert r.output == "file.yaml"
-
-    r = p.parse_args(["compile", "-o", "file.yaml", "foo.yaml"])
-    assert r.command == "compile"
-    assert r.input == "foo.yaml"
-    assert r.output == "file.yaml"
+    if sys_exit_code:
```
mvo5, line 35: (super nitpick) Given how small the body is, I wonder if it's worthwhile to split the test into two, one "happy" and one "sad" case. This way the inputs are slightly smaller and the pytest.mark.parametriz might be slightly easier to read (or maybe not) - in any case, more of a thought then aynthing else, fine as is.

test/test_compile.py
```diff
@@ -0,0 +1,97 @@
+import argparse
+import os.path
+import pytest
+
+from otk.command import compile
+
+
+@pytest.mark.parametrize(
+    "parser, arguments, dry_run, input_data, output_data, sys_exit_code, log_message",
```
mvo5, line 9: I wonder if this one can be split a bit more, it feels a bit hard to read and the test body has a bunch of "if" conditions inside, could this be split into multiple smaller tests maybe somehow? Maybe a "happy" and "sad" might already help?

test/test_compile.py
```diff
@@ -0,0 +1,97 @@
...
+    [
+        # "GENERATE" is a placeholder to append "tmp_path" from pytest
+        # dry_run must not generate an output
+        (
+            None,
```
mvo5, line 14: This seems to be "None" for all inputs? Or did I just not look careful enough?

test/test_compile.py
```diff
@@ -0,0 +1,97 @@
...
+  "test": 1,
+  "version": "2",
+  "sources": {}
+}""",
+            0,
+            None,
+        ),
+    ],
+)
+def test_validate(
```
mvo5, line 64: This is called "test_validate" but the test is called test_compile.py and the function under test is "compile()" too, is this intentional?

mvo5 (commented):
Thanks for the PR! Lots to like here! I added some questions and ideas inline for your consideration. Note that I only reviewed the code changes (first commit).

test/test_compile.py
schuellerf: Well the parameter itself is not even used in the function - yes, I could drop it from the test for now...

src/otk/command.py
```diff
@@ -116,13 +118,22 @@ def compile(
...
+    if not dry_run:
+        dst.write_text(
+            target_registry.get(kind, CommonTarget)().as_string(spec, tree)
+        )
 
     return 0
 
 
+def compile(parser: argparse.ArgumentParser, arguments: argparse.Namespace) -> int:
+    return _process(parser, arguments, False)
```
mvo5, line 129: Maybe change this to `return _process(parser, arguments, dry_runFalse)` (and similar below)? This way it becomes easier to immediately see what the "true/false" mean in here?

test/test_compile.py
schuellerf: @supakeen I can as well remove the parameter "parser" from `compile` (and `validate` and `_process()` how it is called here)
Probably this will never be used?

test/test_compile.py
schuellerf: Nope - just started with validate tests and forgot to rename when I saw we could as well test both

test/test_compile.py
schuellerf: @mvo5 btw. also `arguments.input` is always `"GENERATE"` because I wanted to test the stdin case - but I don't know how to mock `/dev/stdin`. I could mock `sys.stdin` but not `/dev/stdin` - any ideas? 🤔
(without rewriting the whole functionality of otk 😅)

supakeen: Ah yes, I added the `parser` argument I think because I had ideas about testing but they weren't very good. Let's remove that argument.

Perhaps there is a better way than my POSIX extension hacks to do the stdin and stdout handling. I did it this way so the code could just use `pathlib` everywhere. It'd be possible I guess to go with file-descriptors instead.

schuellerf: If you don't mind I'll also change this within this PR as this would be needed to mock stdin I guess

mvo5: Using sys.{stdin,stdout} seems fine with me, is there any downside? It's also easy to mock via pytest

test/test_compile.py
```diff
@@ -0,0 +1,57 @@
...
+  "test": 1,
+  "version": "2",
+  "sources": {}
+}""",
+            0,
+            None,
+        ),
+    ],
+)
+def test_compile(
```
mvo5, line 27: Probably nice to remove the sys_exit_code and log_message as they are currently unused. and/or add a negative test that just makes sure validate fails and does not leave an (invalid/empty) output file behind.

test/test_compile.py
```diff
@@ -0,0 +1,57 @@
...
+    caplog,
+    capsys,
+    monkeypatch,
+    arguments,
+    input_data,
+    output_data,
+    sys_exit_code,
+    log_message,
+):
+    if arguments.input == "GENERATE":
```
mvo5, line 38: (super nitpick) I wonder if the test parameter really needs to be a full argparse.Namespace, it seems we are only intrested in "file" or "stdin/stdout", maybe a combinartion. So just passing "input", "output" as either "some-filename" or "-" (for stdin/stdout) might be slightly shorter and then here we would construct the namespace

test/test_compile.py
```diff
@@ -0,0 +1,56 @@
...
+        input_filename = "input.yaml"
+        arguments.input = os.path.join(tmp_path, input_filename)
```
mvo5, line 40: (super niptick) I like the way pathlib supports joins for paths ` arguments.input = tmp_path / input_filename` (but probably very personal and maybe does not work because input expects a string not a pathlib.Path?

mvo5 (commented):
Thank you! I love how this looks nicer and nicer :) Some more ideas/suggestions (sorry, stop me if you feel it's too much)

mvo5 (approved):
Sorry, I meant to already approve in my previous comment .There are some ideas/suggestions left but it's IMHO a really nice step forwoard and we can do more in followups.

test/test_compile.py
mvo5: Let's keep as is for now and do the stdin/stdout testing in a followup. This way we can also rework it a bit, I think using filedescriptors instead of path has some benefits (including that it will make the sys.std{in,out} tests simpler and that tit will have less TOCTOU issues (which are not very important here)

schuellerf: I don't think this would be TOCTOU...
please check my suggestion (next force push)

test/test_compile.py
schuellerf: oh - yea - I'm just no used to this pathlib feature and still feels strange to override operators like this but I'll change

supakeen (approved):

PR #107, opened 2024-06-05
otk: use file handles instead of paths when reading files

mvo5:
The upside of using file descriptors is that there are less TOCTOU issues, i.e. `if not src.exists()` on a path is racy but this can easily be avoided by just opening the file and keeping it open.

It will also make testing for stdin/stdout easier from the test POV there is nothing special about them (i.e. it's not even needed to test that they work just that when "-" is passed on the commandline (or nothing) stdin/stdout is opened).

Note that there is probably more needed but this is enough to satify the tests (but I suspect the recursive loading will break with this change so it might be a good place to add a test with an `otk.include` to ensure this is covered.

### DIFF
--- a/src/otk/command.py
+++ b/src/otk/command.py
@@ -48,23 +48,18 @@ def root() -> int:
 
 
 def _process(arguments: argparse.Namespace, dry_run: bool) -> int:
-    src = pathlib.Path("/dev/stdin" if arguments.input is None else arguments.input)
-    dst = pathlib.Path("/dev/stdout" if arguments.output is None else arguments.output) if not dry_run else None
-
-    if not src.exists():
-        log.fatal("INPUT path %r does not exist", str(src))
-        return 1
-
-    if not dry_run and not dst.parent.exists():
-        log.fatal("OUTPUT path %r does not exist", str(dst))
-        return 1
+    src = sys.stdin if arguments.input is None else open(arguments.input)
+    if not dry_run:
+        dst = sys.stdout if arguments.output is None else open(arguments.output, "w")
 
     # the working directory is either the current directory for stdin or the
     # directory the omnifest is located in
-    cwd = pathlib.Path.cwd() if arguments.input is None else src.parent
+    cwd = (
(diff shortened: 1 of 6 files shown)


schuellerf (dismissed):
LGTM - very nice - will you add some example tests with sys.stdin/out mocked? otherwise I'd also be happy to take this over

mvo5:
Sure, happy to work on this more and also update the tests, I wanted to double check first if I overlooked something with the idea, I will work on this a bit more and undraft it in a bit.

src/otk/command.py
```diff
@@ -48,27 +48,18 @@ def root() -> int:
...
-        log.fatal("OUTPUT path %r does not exist", str(dst))
-        return 1
+    src = sys.stdin if arguments.input is None else open(arguments.input)
+    if not dry_run:
+        dst = sys.stdout if arguments.output is None else open(arguments.output, "w")
 
     # the working directory is either the current directory for stdin or the
     # directory the omnifest is located in
-    cwd = pathlib.Path.cwd() if arguments.input is None else src.parent
+    cwd = (
```
mvo5, line 57: I really don't get why `ruff` puts this into a multi-line construct with a `(` around but above for `dst = ` it does not do that. I'm really not a fan :(

schuellerf: It's because this line is > 88 chars while `dst` is not.
we could add a `.ruff.toml` with
```
line-length = 120
```

schuellerf: @supakeen any preference for line lengths in this project? I think nowadays with those widescreen, high-res. monitors 120 chars seem fine for me

mvo5: Fwiw, I was a bit grumpy when I wrote it - it's okay if we keep it as is, I'm just not a fan :) [as an additonal datapoint, osbuild uses 120 for the line length]

supakeen: I have no strong opinions; 120 is consistent with `osbuild` so let's do that.

mvo5: Thank you! I opened https://github.com/osbuild/otk/pull/108

supakeen (approved):

PR #110, opened 2024-06-06
otk_osbuild: Override default osbuild-depsolve-dnf with OSBUILD_DEPSOLVE

bcl:
This lets you point to a different location than the default for the depsolver script and makes it easier to experiment with uninstalled versions of osbuild.

Set OSBUILD_DEPSOLVE to the full path of the replacement script or wrapper:

OSBUILD_DEPSOLVE=$HOME/bin/depsolve-wrapper otk compile example/fedora/minimal-40-x86_64.yaml
    
Where the depsolve-wrapper looks like this:
    
PYTHONPATH=/home/user/osbuild/osbuild/ /home/user/osbuild/tools/osbuild-depsolve-dnf $@

### DIFF
--- a/src/otk_osbuild/command.py
+++ b/src/otk_osbuild/command.py
@@ -7,6 +7,7 @@ import sys
 import json
 import hashlib
 import base64
+import os
 import pathlib
 import subprocess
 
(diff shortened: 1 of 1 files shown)


bcl:
I see you're working on an external support system with `OTK_EXTERNAL_PATH` which should handle this.

PR #111, opened 2024-06-07
test: baseline tests

supakeen:
These examples show simple variants of omnifests, not set up to create a full distribution but only to show functionality.

More complex examples are not working at this point; they will come at a later point as the implementation grows to support them.

### DIFF
--- a/test/test_example.py
+++ b/test/test_example.py
@@ -1,26 +1,20 @@
-import pytest
-import pathlib
+import argparse
 import json
+import pathlib
 
-from otk.context import CommonContext, OSBuildContext
-from otk.document import Omnifest
-from otk.transform import resolve
-from otk.target import OSBuildTarget
+import pytest
+
+from otk import command
 
 
-@pytest.mark.parametrize("path", (pathlib.Path(__file__).parent / "data/base").glob("*.yaml"))
-def test_base(path):
-    with open(path) as src, open(path.with_suffix(".json")) as dst:
-        ctx = CommonContext(path.parent)
-        doc = Omnifest.from_yaml_file(src)
+@pytest.mark.parametrize("src_yaml", (pathlib.Path(__file__).parent / "data/base").glob("*.yaml"))
+def test_command_compile_on_base_examples(tmp_path, src_yaml):
+    dst = tmp_path / "out.json"
 
-        tree0 = resolve(ctx, doc.tree)
+    ns = argparse.Namespace(input=src_yaml, output=dst, target="osbuild")
 
-        spc = OSBuildContext(ctx)
-        tree0 = resolve(spc, tree0["otk.target.osbuild.name"])
-        txt = OSBuildTarget().as_string(spc, tree0)
+    command.compile(ns)
 
-        tree0 = json.loads(txt)
-        tree1 = json.load(dst)
-
-        assert tree0 == tree1
+    expected = json.load(src_yaml.with_suffix(".json").open())
+    actual = json.load(dst.open())
+    assert expected == actual
--- a/test/test_example.py
+++ b/test/test_example.py
@@ -1,0 +1,26 @@
+import pytest
+import pathlib
+import json
+
+from otk.context import CommonContext, OSBuildContext
+from otk.document import Omnifest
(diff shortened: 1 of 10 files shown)


example/centos/centos-9-x86_64-ami.yaml
```diff
@@ -35,3 +35,6 @@ otk.define.centos-9-x86_64-ami:
       total_size: 10G
 
 otk.include.centos-9-x86_64: "common/centos-9.yaml"
+
+otk.target.osbuild.name:
```
mvo5, line 39: Is this a leftover?

supakeen: Yea, dangling leftover. I've removed it.

test/test_example.py
```diff
@@ -0,0 +1,26 @@
...
+
+from otk.context import CommonContext, OSBuildContext
+from otk.document import Omnifest
+from otk.transform import resolve
+from otk.target import OSBuildTarget
+
+
+@pytest.mark.parametrize("path", (pathlib.Path(__file__).parent / "data/base").glob("*.yaml"))
+def test_base(path):
+    with open(path) as src, open(path.with_suffix(".json")) as dst:
```
mvo5, line 13: I wonder if this could be written in a slightly more high-level way, this test is using a fairly low-level api but what we want is essentially to test if "compile()" works, so maybe we could use that, something like:
```python
@pytest.mark.parametrize("src_yaml", (pathlib.Path(__file__).parent / "data/base").glob("*.yaml"))
def test_compile_base(tmp_path, src_yaml):
    dst = tmp_path / "out.json"
    # XXX: fix signature of compile
    ns = argparse.Namespace(input=src_yaml, output=dst, target="osbuild")
    otk.command.compile(ns)

    expected = json.load(src_yaml.with_suffix(".json").open())
    actual = json.load(dst.open())
    assert expected == actual
```
It might also be nice to tweak the signature of compile to make it a bit more targeted (as that seems to be all it needs) to something like  `compile(input_path, output_path, target: str)` 

(full diff:
```python
diff --git a/test/test_example.py b/test/test_example.py
index 846d8ff..4ebec6f 100644
--- a/test/test_example.py
+++ b/test/test_example.py
@@ -1,26 +1,22 @@
+import argparse
+import os
 import pytest
 import pathlib
 import json
 
-from otk.context import CommonContext, OSBuildContext
+import otk.command
 from otk.document import Omnifest
 from otk.transform import resolve
-from otk.target import OSBuildTarget
+from otk.target import CommonTarget,OSBuildTarget
 
 
-@pytest.mark.parametrize("path", (pathlib.Path(__file__).parent / "data/base").glob("*.yaml"))
-def test_base(path):
-    with open(path) as src, open(path.with_suffix(".json")) as dst:
-        ctx = CommonContext(path.parent)
-        doc = Omnifest.from_yaml_file(src)
+@pytest.mark.parametrize("src_yaml", (pathlib.Path(__file__).parent / "data/base").glob("*.yaml"))
+def test_compile_base(tmp_path, src_yaml):
+    dst = tmp_path / "out.json"
+    # XXX: fix signature of compile
+    ns = argparse.Namespace(input=src_yaml, output=dst, target="osbuild")
+    otk.command.compile(ns)
 
-        tree0 = resolve(ctx, doc.tree)
-
-        spc = OSBuildContext(ctx)
-        tree0 = resolve(spc, tree0["otk.target.osbuild.name"])
-        txt = OSBuildTarget().as_string(spc, tree0)
-
-        tree0 = json.loads(txt)
-        tree1 = json.load(dst)
-
-        assert tree0 == tree1
+    expected = json.load(src_yaml.with_suffix(".json").open())
+    actual = json.load(dst.open())
+    assert expected == actual
`` )

mvo5 (commented):
Thank you! I like the examples a lot. I made a suggestion in the actual test code, hope it's useful.

test/test_example.py
supakeen: I've pushed this change; for now I'd like to keep the `compile` signature as-is but the rest has been applied.

mvo5 (approved):
Thank you!

PR #113, opened 2024-06-16
otk: some tweaks in the process() recurision

mvo5:
This is a first set of tweaks to make the recursive process() function a bit easier to follow and it also adds a new `ParserState` that is passed inside the recursion and updated as needed. For now it only carries the current active file to track includes and to allow includes relative to the current file that is being worked on. But it could also be used to keep track of e.g. if the parser is currently inside a `otk.define` context (if/when the need for this arises).

Draft because I would like to tweak/add some tests still.

### DIFF
--- a/src/otk/command.py
+++ b/src/otk/command.py
@@ -8,6 +8,7 @@ from .context import CommonContext
 from .context import registry as context_registry
 from .document import Omnifest
 from .help.log import JSONSequenceHandler
+from .parserstate import ParserState
 from .target import CommonTarget
 from .target import registry as target_registry
 from .transform import resolve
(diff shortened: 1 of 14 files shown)


supakeen:
Awesome, we do need something like this. I think the change to make includes relative to the current file instead of relative to the entrypoint is nice too, makes things more re-usable.

mvo5:
> Awesome, we do need something like this. I think the change to make includes relative to the current file instead of relative to the entrypoint is nice too, makes things more re-usable.

Thank you! I forgot to mention in my original description but I think the `ParserState` will also help us detect cyclic includes more easily. There are some more tweaks from the train in the recurison, will they collide with your work in this area or is it okay to open them?

supakeen:
> > Awesome, we do need something like this. I think the change to make includes relative to the current file instead of relative to the entrypoint is nice too, makes things more re-usable.
> 
> Thank you! I forgot to mention in my original description but I think the `ParserState` will also help us detect cyclic includes more easily. There are some more tweaks from the train in the recurison, will they collide with your work in this area or is it okay to open them?

No, and if they do I can incorporate the changes.

mvo5:
I will close this one as @achilleas-k is working on an improved version of this. There might be some testcases that are worth extracting still but keeping this PR open seems unnecessary

PR #115, opened 2024-06-19
test/base: add failing testcases

supakeen:
This PR aims to add many (failing) testcases that should pass to ensure that `otk` implements the entire specification.


test/data/base/03-include-in-define.yaml
```diff
@@ -0,0 +1,9 @@
+otk.version: "1"
+
+otk.define.name:
+  otk.include: "03-include-in-define/example.yaml"
```
mvo5, line 4: I think it's probably a good idea to put this under a subtree to test that this also works, something like:
```yaml
diff --git a/test/data/base/03-include-in-define.yaml b/test/data/base/03-include-in-define.yaml
index f750223..1f6dece 100644
--- a/test/data/base/03-include-in-define.yaml
+++ b/test/data/base/03-include-in-define.yaml
@@ -1,8 +1,9 @@
 otk.version: "1"
 
-otk.define.name:
-  otk.include: "03-include-in-define/example.yaml"
+otk.define:
+  subkey:
+    otk.include: "03-include-in-define/example.yaml"
 
 pipelines:
-  - foo: ${foo}
-    bar: ${bar}
+  - foo: ${subkey.foo}
+    bar: ${subkey.bar}
```

test/data/base/04-op-in-define.yaml
```diff
@@ -0,0 +1,12 @@
+otk.version: "1"
+
+otk.define.name:
+  foo:
+    otk.op.join:
```
mvo5, line 5: A test with an `otk.join` outside a define would be nice too, maybe something like:
```yaml
diff --cc test/data/base/04-op-in-define.yaml
index 9a1143c,1f80deb..0000000
--- a/test/data/base/04-op-in-define.yaml
+++ b/test/data/base/04-op-in-define.yaml
@@@ -7,5 -7,12 +7,11 @@@ otk.define.name
          - - 1
          - - 2
  
- pipelines:
-   - foo: ${foo}
+ otk.target.osbuild.name:
+   pipelines:
+     - foo: ${foo}
 -
+   bar:
+     otk.join:
+       values:
+         - [3, 2, 1]
+         - [6, 5, 4]
diff --git a/test/data/base/04-op-in-define.json b/test/data/base/04-op-in-define.json
index dfab33c..afd6a8a 100644
--- a/test/data/base/04-op-in-define.json
+++ b/test/data/base/04-op-in-define.json
@@ -6,5 +6,8 @@
         2
       ]
     }
+  ],
+  "bar": [
+    1, 2, 3, 4, 5, 6
   ]
 }
```
(I hope this translates right)

test/data/base/04-op-in-define.yaml
```diff
@@ -0,0 +1,12 @@
...
+      values:
+        - - 1
+        - - 2
```
mvo5, line 8: And maybe even more nesting here, a otk.join inside another otk.join (probably a new testcase).

mvo5 (commented):
Thank you! This is nice, I made some inline suggestion and I'm curious why the tests seem to be not failing in this PR, it is marked as "green" for me?

test/data/base/03-include-in-define.yaml
supakeen: I've added a subkey in a separate commit, both an include at the top-level under `otk.define` and the subkey.

test/data/base/04-op-in-define.yaml
supakeen: I've added this as a separate test case (`09-op.yaml`).

test/data/base/04-op-in-define.yaml
supakeen: I added this in the new `09-op` testcase as well.

test/data/base/03-include-in-define/example.yaml
```diff
@@ -0,0 +1,2 @@
+foo: "bar"
+bar: "foo"
```
mvo5, line 2: One more idea, inside the example we should do define a var based on a previous define, both in the simple and in the include case, something like:
```yaml
diff --git a/test/data/base/01-define.json b/test/data/base/01-define.json
index dd2ca6c..a42fb29 100644
--- a/test/data/base/01-define.json
+++ b/test/data/base/01-define.json
@@ -5,6 +5,9 @@
     },
     {
       "a": "tree/b"
+    },
+    {
+      "b": "tree"
     }
   ],
   "version": "2",
diff --git a/test/data/base/01-define.yaml b/test/data/base/01-define.yaml
index 12bb08d..f63e9e8 100644
--- a/test/data/base/01-define.yaml
+++ b/test/data/base/01-define.yaml
@@ -5,8 +5,10 @@ otk.version: "1"
 
 otk.define.name:
   a: "tree"
+  b: ${a}
 
 otk.target.osbuild.name:
   pipelines:
     - a: "${a}"
     - a: "${a}/b"
+    - b: "${b}"
diff --git a/test/data/base/03-include-in-define.json b/test/data/base/03-include-in-define.json
index d84def4..09d8985 100644
--- a/test/data/base/03-include-in-define.json
+++ b/test/data/base/03-include-in-define.json
@@ -2,7 +2,8 @@
   "pipelines": [
     {
       "foo": "bar",
-      "bar": "foo"
+      "bar": "foo",
+      "baz": "bar"
     }
   ],
   "version": "2",
diff --git a/test/data/base/03-include-in-define.yaml b/test/data/base/03-include-in-define.yaml
index f0cc68c..6a155b5 100644
--- a/test/data/base/03-include-in-define.yaml
+++ b/test/data/base/03-include-in-define.yaml
@@ -8,3 +8,4 @@ otk.target.osbuild.name:
   pipelines:
     - foo: ${subkey.foo}
       bar: ${subkey.bar}
+      baz: $subkey.baz}
diff --git a/test/data/base/03-include-in-define/example.yaml b/test/data/base/03-include-in-define/example.yaml
index bba4346..12c7125 100644
--- a/test/data/base/03-include-in-define/example.yaml
+++ b/test/data/base/03-include-in-define/example.yaml
@@ -1,2 +1,3 @@
 foo: "bar"
 bar: "foo"
+baz: ${foo}
```
(please double check). This will ensure that the parser "knows" when inside a define in an include, i.e. that it needs to keep track of newly includes vars inside a define as well. Maybe even add a dict here (baz:\n  subkey: something) so that we can be sure this part of the recursion is also working.

mvo5 (commented):
One more quick idea during my train ride, also do we allow otk.join inside otk.define (I don't remember)?

test/data/base/03-include-in-define/example.yaml
achilleas-k: I'm okay to merge this PR as is and we can PR this example along with other ones I have in mind.

mvo5: Sure, that works for me of course

achilleas-k (approved):

PR #117, opened 2024-06-21
gh: add workflow to run tests as part of each PR

mvo5:
Small workflow so that we run the unit tests as part of our PR workflow.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,30 @@
+name: Test
+# Make sure only one action triggers the job, otherwise pushing to a
+# pull-request will run it twice.
+on:
+  pull_request:
+    branches:
+      - "*"
+  merge_group:
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -0,0 +1,31 @@
...
+      - name: "Clone Repository"
+        uses: actions/checkout@v4
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v5
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          # XXX: install from pyproject.toml
```
supakeen, line 28: Do you want to address that in this PR?

mvo5: Sorry, yes! This is hopefully fixed now in the right way.

supakeen: It is :)

supakeen (approved):

PR #119, opened 2024-06-21
github/workflows: fix false positive yamllint message

schuellerf:
The format of github confuses yamllint, so we'll ignore the message here


mvo5 (approved):
Thank you!

supakeen (approved):

schuellerf:
also I set pre-commit to be as strict as GH so we see it locally

PR #120, opened 2024-06-21
Small refactors and cleanups in prep for bigger changes

achilleas-k:
This PR includes some small refactoring, code cleanups, and a new test.
I also included the first commit from #113 to avoid conflicts.  The second commit I'd like to add later with some changes (for example c6b0b6871b4928ed4b8b1df0b23728a0752821cc).

I renamed the `desugar()` function for clarity.  Quoting the commit message:

> The name `desugar()` was in reference to initial specifications that supported multiple ways of referring to variables and the `${name}` form was considered "syntactic sugar".  Now we only have one way of referring to variables and the concept of "desugaring" might be hard to understand.
>
> Rename it to `replace_vars()` to make its purpose clearer.

### DIFF
--- a/src/otk/directive.py
+++ b/src/otk/directive.py
@@ -76,7 +76,7 @@ def define(ctx: Context, tree: Any) -> Any:
 def include(ctx: Context, tree: Any) -> Any:
     """Include a separate file."""
 
-    tree = desugar(ctx, tree)
+    tree = substitute_vars(ctx, tree)
 
     file = ctx._path / pathlib.Path(tree)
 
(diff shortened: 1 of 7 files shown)


src/otk/directive.py
```diff
@@ -148,52 +148,56 @@ def _op_map_join(ctx: Context, values: List[dict]) -> Any:
...
+            if isinstance(value, (int, float)):
+                value = str(value)
 
             # Any other type we do not
-            if not isinstance(data, str):
+            if not isinstance(value, str):
                 raise TransformDirectiveTypeError(
-                    "string sugar resolves to an incorrect type, expected int, float, or str but got %r",
-                    data,
+                    f"string {data} resolves to an incorrect type, "
```
supakeen, line 189: Let's make it `{data!r}`.

supakeen (approved):

src/otk/directive.py
achilleas-k: Right, good idea.  Done.

supakeen (approved):

PR #123, opened 2024-06-21
doc: trivial update for `desugar` -> substitution

mvo5:
Trivial followup for PR#120

### DIFF
--- a/doc/03-omnifest/01-directive.md
+++ b/doc/03-omnifest/01-directive.md
@@ -73,7 +73,7 @@ otk.include: ${variable}
 ```
 
 If a `${}` appears in a `str` value then its string value as it appears
-in `otk.define` is replaced into the string. Note that using the sugared form
+in `otk.define` is replaced into the string. Note that substitutions
 in this form requires the value to be a string in the `otk.define`.
 
 ```yaml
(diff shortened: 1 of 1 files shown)


achilleas-k (approved):

doc/03-omnifest/01-directive.md
```diff
@@ -73,7 +73,7 @@ otk.include: ${variable}
 ```
 
 If a `${}` appears in a `str` value then its string value as it appears
-in `otk.define` is replaced into the string. Note that using the sugared form
+in `otk.define` is replaced into the string. Note that substitutions
 in this form requires the value to be a string in the `otk.define`.
```
achilleas-k, line 77:
```suggestion
in `otk.define` is replaced into the string. Note that substitutions
in this form require the value to be a string in the `otk.define`.
```

achilleas-k (changes requested):
Approved too fast.

doc/03-omnifest/01-directive.md
mvo5: Hm, hm, I'm not a native speak but to me both sound okay? Fine committing this though (as both sound okay to me :)

achilleas-k (approved):
Typo in the second commit message, but I'm good to merge if you are.

PR #125, opened 2024-06-21
Fix precommit yamllint

schuellerf:
as `pre-commit` only runs on files you currently changed by default `lint.yml` and `test.yml` have been missing...

also suggesting to use autopep8 as it's not that aggressive as `ruff` is (see discussion in https://github.com/osbuild/images/pull/753 )


supakeen (approved):

supakeen (changes requested):
`ruff` needs to be dropped from dev dependencies and `autopep8` added.

Issue #112, opened 2024-06-12
`otk.target` aren't recognized if defined inside `otk.include`


ondrejbudai:
`a.yaml`:

```
otk.version: '1'

otk.include.b: "b.yaml"

```

`b.yaml`:

```
otk.target.foo.bar:
  yolo: true
```

```
python3 -m otk validate ../a.yaml
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/home/obudai/work/otk/src/otk/__main__.py", line 3, in <module>
    root()
  File "/home/obudai/work/otk/src/otk/command.py", line 45, in root
    return validate(arguments)
           ^^^^^^^^^^^^^^^^^^^
  File "/home/obudai/work/otk/src/otk/command.py", line 118, in validate
    return _process(arguments, dry_run=True)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/obudai/work/otk/src/otk/command.py", line 60, in _process
    doc = Omnifest.from_yaml_file(src)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/obudai/work/otk/src/otk/document.py", line 33, in from_yaml_file
    return cls.from_yaml_bytes(file.read())
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/obudai/work/otk/src/otk/document.py", line 25, in from_yaml_bytes
    cls.ensure(deserialized_data)
  File "/home/obudai/work/otk/src/otk/document.py", line 64, in ensure
    raise ParseTargetError("omnifest must contain at least one key by the name of `%s.*`" % (PREFIX_TARGET,))
otk.error.ParseTargetError: omnifest must contain at least one key by the name of `otk.target..*`
```

supakeen:
Yea, we should fix this. We should probably also change the entrypoint documentation to allow `otk.target` via includes.

PR #126, opened 2024-06-21
Refactor and fix up recursive resolve and processing functions

achilleas-k:
This PR refactors and fixes the recursive resolve functionality to match the spec.  All but one of the examples in `test/data/base/` pass successfully.  The `05-target-in-include.yaml` example fails because the `_process()` function in `command.py` expects a target in the entrypoint and errors if it doesn't find one.

Commit messages should tell the whole story, but let me highlight a few things here I think are the most important:
- The new ParserState, originally introduced by @mvo5 in another PR and pulled in here, tracks two things: the path to the file being processed and a reference to the global defines or a sub-part of them depending on the state.  When a sub-block of a defines block is being processed, the `ParserState.defines` references the sub-block being processed, so variable definitions can be added to that sub-block directly.  It will also, eventually, probably carry a stack of directives being processed so we can do things like disallow includes under defines.
- The `process_include()` function loads a yaml file and sends it through `resolve()` before returning the contents.  I think this should eventually be the entrypoint for the recursion.  In other words, the path to the entrypoint yaml file specified on the command line can be sent directly through to `process_include()` and it should all work out.
    - However, this means that we wont be able to (nor I think we would have to) do any preprocessing of the entrypoint.  I think eventually it might be good to load the file, start processing, and validate after everything is included and resolved.
- Target processing is currently ignored.  I don't think this is particularly difficult to take care of.  I think a good approach to target processing and selection might be to process all targets found while running through the `resolve()` recursion, then print the target that matches the one specified on the command line (or the default if there's only one).  Before printing we would have a tree that contains everything we loaded and resolved, so we could print the flattened document with all targets included and variables resolved and we could also run some validation over it.
- The `process_defines()` is different from all the other ones in that it modifies the context in-place and doesn't return anything.  I wonder if things would be simpler if every resolver/process function worked this way and modified the tree in-place.


achilleas-k:
I just realised that the only failing test now is the `target-in-include` test that fails during the initial validation that requires a target in the entrypoint.

supakeen:
> I just realised that the only failing test now is the `target-in-include` test that fails during the initial validation that requires a target in the entrypoint.

Feel free to drop it in this PR if that turns it green; we decided in #112 to do so.

achilleas-k:
Disabled test (without removing it) and marked as ready.

src/otk/command.py
```diff
@@ -8,6 +8,7 @@
 from .context import registry as context_registry
 from .document import Omnifest
 from .help.log import JSONSequenceHandler
+from .parserstate import ParserState
```
supakeen, line 11: I don't want to do naming discussions but I also can't not say this: it isn't the state of the parser, `traversal.State` or such would make more sense to me. Feel absolutely free to ignore.

src/otk/parserstate.py
```diff
@@ -0,0 +1,17 @@
+class ParserState:
+
+    def __init__(self, path, defines):
+        self.path = path
+        self.defines = defines
+
+    def new(self, *, path=None, defines=None) -> "ParserState":
+        """
+        Return a new ParserState, optionally redefining the path and defines properties.
+        Properties not defined in the args are (shallow) copied from the existing instance.
```
supakeen, line 10: Could we put the 'why' here as well?

src/otk/parserstate.py
```diff
@@ -0,0 +1,17 @@
+class ParserState:
+
+    def __init__(self, path, defines):
+        self.path = path
+        self.defines = defines
+
+    def new(self, *, path=None, defines=None) -> "ParserState":
```
supakeen, line 7: Since this is all pretty close to `contextvar`'s, perhaps we can call this `def copy(...)`?

supakeen (commented):
Really just naming things otherwise great stuff <3

src/otk/command.py
mvo5: Thanks, that is a good point, ParserState may not be ideal, `{traversal,recursion,rec,trans}.State` are some alternative ideas (fwiw, I don't have a super strong opinion here)

achilleas-k: I renamed the module to `traversal` and the class to `State`.

src/otk/parserstate.py
achilleas-k: Added the following to the docstring:
```
        A shallow copy of the 'defines' allows us to keep modifying the global
        context variables through the reference on this State object, which can
        refer to a sub-object of the global context when necessary.
```

src/otk/parserstate.py
achilleas-k: Renamed to `State.new()`

supakeen (approved):

src/otk/traversal.py
```diff
@@ -1,3 +1,20 @@
 class State:
-    def __init__(self, path):
+    def __init__(self, path, defines):
         self.path = path
+        self.defines = defines
+
+    def copy(self, *, path=None, defines=None) -> "State":
```
mvo5, line 6: (super nitpick) the commit message talks about `new()` but this is now `copy()`

src/otk/context.py
```diff
@@ -8,13 +8,10 @@
...
 from typing import Any, Optional
 
-from .error import (
-    TransformDefineDuplicateError,
-    TransformVariableIndexRangeError,
-    TransformVariableIndexTypeError,
-    TransformVariableLookupError,
-    TransformVariableTypeError,
-)
+from .error import (TransformDefineDuplicateError,
```
mvo5, line 11: (super nitpick) I wonder if this `otk: isort context.py`  commit could be squashed into some other commit?

src/otk/transform.py
```diff
@@ -112,15 +112,45 @@ def is_directive(needle: Any) -> bool:
...
+
+    # Iterate over a copy of the tree so that we can modify it in-place.
+    for key, value in tree.copy().items():
+        if key.startswith("otk.define"):
+            # nested otk.define: process the nested values directly
+            process_defines(ctx, state, value)
+            continue
+
+        if key.startswith("otk.include"):
+            # TODO: disallow this (see https://github.com/osbuild/otk/issues/116)
```
mvo5, line 144: I guess the plan here is that we introduce it and then revert in a single commit so that (if we actually change our mind) we can get it back trivially ? Did I get this right?

src/otk/transform.py
```diff
@@ -112,15 +112,45 @@ def is_directive(needle: Any) -> bool:
...
+    and update the global context. The State holds a reference to the nested
+    defines block the function is working in. New defines are added to the
+    nested block but references are resolved from the global ctx.defines.
+    """
 
-    return tree
+    subblock = state.defines
+
+    # Iterate over a copy of the tree so that we can modify it in-place.
+    for key, value in tree.copy().items():
```
mvo5, line 137: The commit message for this new recursive function has some very nice examples - I wonder if those could be (easily?) added as test-cases here? Having this commit without tests makes me slightly uneasy (I know tests are coming later/are part of the pre-req but I wonder it it still might be useful to illustrate the new code?)

[edit: tests are in a separate commit, I guess that is fine and just personal preference (to me they are kinda the same workstream so having them in the same commit seems fine but I don't mind to much either way)]

src/otk/transform.py
```diff
@@ -65,31 +65,37 @@ def resolve_dict(ctx: Context, state: State, tree: dict[str, Any]) -> Any:
         if isinstance(val, str):
             val = substitute_vars(ctx, val)
         if is_directive(key):
-            # Define, target, and version are done separately, they allow
-            # sibling elements thus they return the tree with their key set to their
-            # (resolved) value.
+
```
mvo5, line 68: (super nitpick) stray newline here (or is it intensional?)

src/otk/transform.py
```diff
@@ -59,6 +59,11 @@ def resolve_dict(ctx: Context, state: State, tree: dict[str, Any]) -> Any:
     """
 
     for key, val in tree.items():
+        # Replace any variables in a value immediately before doing anything
+        # else, so that variables defined in strings are considered in the
+        # processing of all directives.
+        if isinstance(val, str):
```
mvo5, line 65: I wonder if a test for this could be easily extracted from the example given in the commit message for this change?

test/data/error/00-invalid.err
```diff
@@ -0,0 +1 @@
+00-invalid.yaml
```
mvo5, line 1: Maybe expand this a little bit more, something like `00-invalid.yaml was not found` maybe?

mvo5 (dismissed):
Thank you! This is very nice indeed - I have a bunch of small nitpicks/ideas/suggestions but if we want to move fast we can ignore them all and just merge as is (I went over this commit by commit so some comments are just related to the commit order/granularity).

src/otk/traversal.py
achilleas-k: oh no, I missed that!

src/otk/context.py
achilleas-k: I'd rather keep codestyle / formatting changes in separate commits so it doesn't look like that imports changed in a commit that has nothing to do with import changes.

src/otk/transform.py
achilleas-k: Yes, mostly.  But also my thinking was that this is a spec change that came from a discussion while the implementation was happening.  I wanted to isolate the change so that it has more visibility and stands on its own (e.g. a PR that only changes this, linked from the issue, that can be read on its own).

Also, I wasn't sure at the time if it needed some state tracking (something to tell us that we're in a defines block).  Now I realise since defines can't branch out to any other directives (except other defines), it's not really necessary.  But we should also think about what happens with other directives in general.

src/otk/transform.py
achilleas-k: It was intentional to make the nested if blocks stand out:
```python
        if is_directive(key):

            if key.startswith(PREFIX_DEFINE):
                process_defines(ctx, state, val)
                continue

            if key == NAME_VERSION:
                continue

            if key.startswith(PREFIX_TARGET):
                continue
```

src/otk/transform.py
achilleas-k: One of the example tests covers this (`09-include-with-var`).  Or would you rather also have a small unit test for it?

test/data/error/00-invalid.err
achilleas-k: Done.

I also changed the test code to catch a generic `Exception`.  Before it was specifically catching `FileNotFoundError` because it was only testing this example, but that would break as soon as we introduce another test.

src/otk/transform.py
achilleas-k: Squashed the two commits.

src/otk/traversal.py
mvo5: It seems it still does(?) (c.f. https://github.com/osbuild/otk/pull/126/commits/d1ab6046f376079c2862016a6b8f954920720b70)

achilleas-k: Fixed.

mvo5 (approved):
Very nice, thank you!

PR #127, opened 2024-06-24
otk: detect circular imports

mvo5:
Now that we have state during the recursion we can detect circular imports. This commit adds it and a matching test.

### DIFF
--- a/src/otk/error.py
+++ b/src/otk/error.py
@@ -98,3 +98,9 @@ class TransformDirectiveUnknownError(TransformError):
     """Unknown directive."""
 
     pass
+
+
+class CircularIncludeError(ParseError):
+    """Cirtcular include detected."""
+
+    pass
--- a/src/otk/transform.py
+++ b/src/otk/transform.py
@@ -13,6 +13,7 @@ from __future__ import annotations
 
 import itertools
 import logging
+import os
 import pathlib
 import re
 from typing import Any, List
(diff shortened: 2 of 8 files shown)


src/otk/transform.py
```diff
@@ -65,7 +65,6 @@ def resolve_dict(ctx: Context, state: State, tree: dict[str, Any]) -> Any:
         if isinstance(val, str):
             val = substitute_vars(ctx, val)
         if is_directive(key):
-
```
mvo5, line 68: This was added by precommit :( (which I still do not like very much!)

src/otk/error.py
```diff
@@ -98,3 +98,9 @@ class TransformDirectiveUnknownError(TransformError):
     """Unknown directive."""
 
     pass
+
+
+class CircularIncludeError(OTKError):
```
supakeen, line 103: Let's inherit from `ParseError`?

src/otk/transform.py
supakeen: Will be fixed after #125.

supakeen (approved):
Tiny typo in a docstring if you want to address it, otherwise great.

src/otk/traversal.py
```diff
@@ -16,5 +19,7 @@ def copy(self, *, path=None, defines=None) -> "State":
             path = self.path
         if defines is None:
             defines = self.defines
+        if includes is None:
+            includes = self.includes.copy()
```
achilleas-k, line 23: Late to the party but:
Should we make `copy()` "smarter" and add something like:
```
if path != includes[-1]:
  includes.append(path)
```
so that whenever the path changes it's automatically added to the include stack?

Too smart?

PR #128, opened 2024-06-24
delay target verification

supakeen:
Delay the target verification until *after* processing includes, fixes #112.

PR #130, opened 2024-06-24
[RFC] use process_include() for main entry point too

mvo5:
A quick draft with some ideas about consolidating some things now that we have a the new recursive core.

---

otk: remove no longer used `include()` and tests

The `transform.include()` helper got superseeded by the new
`process_includes()`.

----

otk: simplify Omnifest and use `process_include` in
 entrypoint

With the recent change to require a fully processed tree before
creating the omnifest [0] it seems we can simplify the Omnifest
a bit more now that it essentially is just a container for an
externally resolved tree.

This commit does that by removing some of the entrypoints
like `from_yaml_{file,data}` as those are not that useful
without resolving the tree first.

One nice effect of this change is that the circular import
detection now includes the top-level entrypoint in the error
and it is also more symetrical (a single way to open/read/parse
an otk file).

The handling of data from <stdin> is not super nice right now
and it actually makes me wonder if we should maybe stop supporting
omnifests from stdin. It seems that most omnifests will contain
includes so the notion of creating an omnifest from <stdin> is
a bit fuzzy.

[0] https://github.com/osbuild/otk/pull/128

---

otk: move validations that targets exist into
 Omnifest.ensure()

This commit moves all the validation for a well-formed omnifest
into the `Omnifest.ensure()` helper.

### DIFF
--- a/src/otk/transform.py
+++ b/src/otk/transform.py
@@ -186,22 +186,6 @@ def process_include(ctx: Context, state: State, path: pathlib.Path) -> dict:
     return {}
 
 
-@tree.must_be(str)
-def include(ctx: Context, tree: Any) -> Any:
-    """Include a separate file."""
-
-    tree = substitute_vars(ctx, tree)
-
-    file = ctx._path / pathlib.Path(tree)
-
-    # TODO str'ed for json log, lets add a serializer for posixpath
-    # TODO instead
-    log.info("otk.include=%s", str(file))
-
-    # TODO
-    return yaml.safe_load(file.read_text())
-
-
 def op(ctx: Context, tree: Any, key: str) -> Any:
     """Dispatch the various `otk.op` directives while handling unknown
     operations."""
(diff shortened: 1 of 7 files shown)


src/otk/transform.py
```diff
@@ -168,10 +168,11 @@ def process_include(ctx: Context, state: State, path: pathlib.Path) -> dict:
     Load a yaml file and send it to resolve() for processing.
     """
     # resolve 'path' relative to 'state.path'
-    cur_path = state.path.parent
-    path = (cur_path / pathlib.Path(path)).resolve()
+    if not path.is_absolute():
+        cur_path = state.path.parent
+        path = (cur_path / pathlib.Path(path)).resolve()
```
achilleas-k, line 173: :heart:

achilleas-k (commented):
Really like the simplifications here.

Looks like there's a lot of left-over unused imports though.
Other than that, lgtm.

supakeen (approved):
Decrufting is the best, thanks.

PR #131, opened 2024-06-24
ci: use pre-commit in actions as well

supakeen:
We dropped `ruff` for formatting in #125, lets drop `ruff` from `make lint`; let's use `pre-commit` in CI for the linter to make everything the same everywhere.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,6 @@
 .PHONY: lint
 lint:
-	@find . -name '*.yaml' | xargs yamllint
-	@ruff check
+	pre-commit run --all-files
 
 .PHONY: type
 type:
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -9,7 +9,8 @@ type:
 
 .PHONY: format
 format:
-	@ruff format src/ test/
+	@find src -name '*.py' | xargs autopep8 --in-place
+	@find test -name '*.py' | xargs autopep8 --in-place
```
achilleas-k, line 12: I ran this against simply doing
```
autopep8 --in-place --recursive .
```
and it's a tiny bit faster (milliseconds).  I wonder if it'll make a difference in the long run when our file numbers grow.

achilleas-k (approved):

Makefile
achilleas-k: Actually, the issue seems to be the double call to find.
```
@find src test -name '*.py' | xargs autopep8 --in-place
```
is also a bit quicker.

supakeen: Went for the second option in a new commit :)

achilleas-k (approved):

PR #134, opened 2024-06-24
otk: consolidate target filtering into `Omnifest.targets`

mvo5:
There are some places that check for targets, consolidate the target filtering into Omnifest and the error checking too.


supakeen (approved):
Looks good. `_targets` doesn't necessarily need a prefix like this; maybe just calling it `filter_targets` would suffice? :)

PR #135, opened 2024-06-24
[RFC] otk: rename `traverse.State.defines` to `define_subtree_ref`

mvo5:
To make clearer what the var is about name it more explicitely.

Small followup for https://github.com/osbuild/otk/pull/129#discussion_r1651263922 - names are a bit strawman(ish) but I feel this would makes it a little bit clearer. 

We could also remove Context.define() as it is no longer used (outside of test code) or we change the code so that the Context.define() is used again but then we will need to support "nested.defines.with.subkeys" :) (but it may make the recursive defines also nicer but is a slightly bigger change)


achilleas-k:
Should we close this in favour of the work you started with #137?

Issue #136, opened 2024-06-25
Make process_defines() easier to follow


achilleas-k:
Comment from @mvo5 in #129:

(nitpick) I was having a bit of a difficult time reading this, maybe just the sun. I wonder if a slightly diffrent way of phrasing it is easier to read (but maybe it's just me):
```python
diff --git a/src/otk/transform.py b/src/otk/transform.py
index fef9428..0e93cfc 100644
--- a/src/otk/transform.py
+++ b/src/otk/transform.py
@@ -148,14 +148,14 @@ def process_defines(ctx: Context, state: State, tree: Any):
             subblock.update(incl)
 
         if isinstance(value, dict):
-            # the value is a dict: process it recursively under a new subblock
-            new_subblock = {}
-            if isinstance(subblock.get(key), dict):
-                # If the new subblock already exists and is a dictionary, use it so they get merged
-                # Any other type will be replaced completely.
-                new_subblock = subblock[key]
-
-            # set the new subblock on the parent so that both context and subblock are available immediately
+            # The value is a dict: process recursively and keep track of
+            # new subblocks to keep track of nested values
+            new_subblock = subblock.get(key)
+            if not isinstance(new_subblock, dict):
+                # if the new block is not a dict it gets replaced completely,
+                # this happens e.g. when redefining vars
+                new_subblock = {}
+
             subblock[key] = new_subblock
             new_state = state.copy(defines=new_subblock)
             process_defines(ctx, new_state, value)
```
*but* I think something slightly deeper is irking me, it seems like the separation between "Context" and "State" is  a bit loose right now and Context._variables is refrenced from state but the connection is not super clear (to me)  - I have no good idea yet but I think this is part of my struggle :)

_Originally posted by @mvo5 in https://github.com/osbuild/otk/pull/129#discussion_r1651263922_
            
(see thread for further discussion)

---

Ideas for addressing this to make it easier to follow:
- Do not attach a reference to a subset of (or the whole) the global defines to the traversal State.  Instead, add a list of keys to the State that tracks which defines block is being processed.
```python
if isinstance(value, dict):
  ...
  new_state = state.copy(defines_key=key)
  process_includes(ctx, new_state, ...)
```

- Add new values to the global context directly, using the keys from the State.
```python
  ctx.defines[state.defines_key] = new_value  # update the context directly
```

- Clearly separate the three cases of interacting value types.
```python
if isinstance(value, dict) and isinstance(ctx.defines.get(key), dict):
  # merge code
[...]

Issue #116, opened 2024-06-20
Nested variable definitions, sub-variable references, and includes


achilleas-k:
I came up with a fun little example of nested variable definitions and references:
```yaml
otk.version: "1"

otk.define.x:
  a: 1
  b:
    a: 2  # b.a
    outer: "${a}"  # 1
    inner: "${b.a}"  # 2


otk.target.osbuild.x:
  vars:
    - "{a}"  # 1
    - "{b}"  # { ... }
    - "{b.a}"  # 2
    - "{b.outer}"  # 1
    - "{b.inner}"  # 2
  ```

There's an interesting thing going on with the context that we need to be explicit about.  When resolving defines, we need to have the global defines available but ONLY for reading.  Because when we're defining the `b:` block of variables in this case, we want to be able to read the global, upper level value of `${a}`, but we should only be able to define keys under `b:`.  So `outer:` shouldn't be set in the global context, but a kind of sub-context that is nested under `b:`.
We could do this with some special handling of the context or we can consider the context "read only" while working on defines and then return the new values from the defines handler and set them externally, under the key we were processing.  But I think that wont work either, because the new defines should be available immediately as well (e.g. `${b.a}` is already defined when processing the value of `outer:`).

It might be easier to have direct access to the defines dict from the context here.  Because then we can do:
```python
new_block = context.defines.get("b", {})  # get the "b" block if it already exists so we can update it
define(context, tree["b"], new_block)  # runs through the "b" subtree populating new_block with new values and using context.defines for resolving references
context.defines["b"] = new_block  # set the new stuff under "b"
```

However, this whole thing brings up another interesting edge case when includes are thrown in the mix:
```yaml
otk.define.first:
  outer:
    a: 1
    otk.include: "other.yaml"

otk.define.second:
  otk.include: "other.yaml"

--- [other.yaml]
otk.define.other:
  a: 2
  b: "${a}"
```

If defines are really global, `b: ${a}` will be set to `1` in the first case and `2` in the second, because the meaning of `${a}` changes.  In the first case, `${a}` refers to the global defined in the first block and the other `a:` gets registered as `${outer.a}`.  In the second case, `a: 2` overrides the global `${a}` because the include is not nested.
This might be a problem because variables in includes will behave differently based on where the define is nested.  When writing an otk fragment, I would expect that:

```yaml
plat: "f38"
plat_long: "platform:${plat}"
```

[...]

PR #137, opened 2024-06-25
conext: add support for dot notation in CommonContext.define()

mvo5:
This adds support for setting dotted values in `CommonContext.define()`. It also removes the `duplicate_definitions_allowed` as this is currently not used.

The removal of `duplicate_definitions_warning` is temporary and it should be added back with a proper test.

PR #139, opened 2024-06-25
otk: use new CommonContext.define() when parsing defines

mvo5:
Based on https://github.com/osbuild/otk/pull/137

One possible way to make the process_define() slightly easier to follow, an alternative idea to https://github.com/osbuild/otk/pull/135

Note that this also removes the ability to use "otk.include" inside "otk.define" as it seemed wasteful to first port the code to the new approach and then delete it again.

Closes https://github.com/osbuild/otk/issues/136
Closes https://github.com/osbuild/otk/issues/116


src/otk/transform.py
```diff
@@ -140,32 +138,27 @@ def process_defines(ctx: Context, state: State, tree: Any):
...
-                # If the new subblock already exists and is a dictionary, use it so they get merged
-                # Any other type will be replaced completely.
-                new_subblock = subblock[key]
-
-            # set the new subblock on the parent so that both context and subblock are available immediately
-            subblock[key] = new_subblock
-            new_state = state.copy(defines=new_subblock)
+            new_define_subkeys = state.define_subkeys.copy()
+            new_define_subkeys.append(key)
+            new_state = state.copy(define_subkeys=new_define_subkeys)
```
achilleas-k, line 153: I think this would become much easier to read if `State` gave us some helper methods for this.  I'm thinking maybe `state.define_subkeys` could return the `"a.b"` form itself and also the `copy()` method could instead do:
```python
def copy(..., subkey_add: str = None):
  new_subkeys = self._subkeys.copy()
  if subkey_add:
    new_subkeys.append(subkey_add)
  return State(..., new_subkeys)
```

and
```python
@property
def define_subkeys(self):
  return ".".join(self._subkeys)
```

achilleas-k (dismissed):
This looks good!  Small comments inline, for potential follow-up.

src/otk/transform.py
mvo5: Thanks for this idea! I reworked the state encapsulation a bit, I hope it captures your suggestion.

supakeen (approved):
Looking nice.

src/otk/traversal.py
```diff
@@ -1,24 +1,52 @@
...
-            define_subkeys = self.define_subkeys
-        if includes is None:
-            includes = self.includes.copy()
+        new_state = copy.deepcopy(self)
+        if subkey_add:
+            new_state._define_subkeys.append(subkey_add)
+        if path:
+            if path in self._includes:
+                circle = [os.fspath(p) for p in new_state._includes]
+                raise CircularIncludeError(f"circular include from {circle}")
```
achilleas-k, line 32: Nitpick: would this be nicer as:
```python
f"circular include: ' -> '.join(circle)
```
and also if `circle.append(path)` so we get:
```
circular include: a.yaml -> b.yaml -> c/c.yaml -> a.yaml
```

achilleas-k (approved):
Very nice.  Love the new State.

src/otk/traversal.py
achilleas-k: Let's merge and do this in another PR if we like it.

PR #140, opened 2024-06-26
context: bring back override warnings

mvo5:
Based on https://github.com/osbuild/otk/pull/139, only the last commit is relevant

---

context: bring back override warnings

This commit brings back the var override warnings, they will probably
be useful for debugging. The format is little funky when nested
vars get replaced that used to be values and need to become dicts.
E.g.
```
key:
 sub: subval
```
and with something like:
```
ctx.define("key.sub.subsub.subsubsub", {"subsubsub": "val2"})
```
the generated warning will generate a doted notation in the
dict which is a bit strange but probably good enough:
```
redefinition of 'key.sub', previous values was 'subval' and new value is {'subsub.subsubsub': {'subsubsub': 'val2'}}
```
The alternative would be to construct the full dict which is not
hard but probably also does not add too much extra clarity(?).


supakeen (approved):
Very nice. I don't think there's another better solution for showing the keys/values.

achilleas-k (approved):
FWIW, I think
```
redefinition of 'key.sub', previous values was 'subval' and new value is {'subsub.subsubsub': {'subsubsub': 'val2'}}
```
doesn't look too bad and certainly tells the whole story, so it's good enough as a warning.

PR #141, opened 2024-06-26
test/base: failing include-in-op

supakeen:
When a file is included in an `otk.op.join` it gets treated as a dictionary however in this case I'm including lists.


schuellerf (approved):
LGTM

src/otk/transform.py
```diff
@@ -77,7 +77,13 @@ def resolve_dict(ctx: Context, state: State, tree: dict[str, Any]) -> Any:
             if key.startswith(PREFIX_INCLUDE):
                 tree = tree.copy()  # copy the tree and drop the include directive
                 del tree[key]
-                tree.update(process_include(ctx, state, pathlib.Path(val)))
+                included = process_include(ctx, state, pathlib.Path(val))
+
+                if isinstance(included, list):
```
mvo5, line 82: I was wondering about non-list types here when reviewing, it got merged before already so I did a quick followup in https://github.com/osbuild/otk/pull/145

PR #143, opened 2024-06-26
otk: limit var naming

mvo5:
This commit adds rules how variables can be named. This is
mostly a defensive measure to avoid ambiguity to disallow
cases that are valid yaml but will may confuse users:
```yaml
otk.define:
 a-b: "foo"

key: ${a-b}
```
which is no longer possible with this commit.

### DIFF
--- a/doc/03-omnifest/01-directive.md
+++ b/doc/03-omnifest/01-directive.md
@@ -60,6 +60,10 @@ otk.define:
   boot_mode: uefi
 ```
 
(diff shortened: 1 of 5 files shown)


supakeen (changes requested):
Looks good however, let's share the pattern with where we do string interpolation: https://github.com/osbuild/otk/blob/main/src/otk/transform.py#L253, I'd prefer the regex that's in *this* PR so let's move the regex to `otk.constant` and then use it in both places?

supakeen (approved):

PR #147, opened 2024-06-27
context: fix error when a variable lookup does not work

mvo5:
The variable lookup in context was missing error information, this commit adds a test and adds the missing error information.

### DIFF
--- a/src/otk/context.py
+++ b/src/otk/context.py
@@ -83,10 +83,10 @@ class CommonContext(Context):
 
         value = self._variables
 
-        for part in parts:
+        for i, part in enumerate(parts):
             if isinstance(value, dict):
                 if part not in value:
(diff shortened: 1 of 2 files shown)


src/otk/context.py
```diff
@@ -83,10 +83,10 @@ def variable(self, name: str) -> Any:
 
         value = self._variables
 
-        for part in parts:
+        for i, part in enumerate(parts):
             if isinstance(value, dict):
                 if part not in value:
-                    raise TransformVariableLookupError(f"Could not find {parts!r}")
+                    raise TransformVariableLookupError(f"could not find {parts!r}")
```
mvo5, line 89: (super nitpick) this is just done for consistency - either all errors should be capitalized or none, I opted for lowecase but I have no strong opinion as long as we are consistent

supakeen (approved):

PR #149, opened 2024-06-27
otk: make code `mypy` clean and run as part of the gh actions

mvo5:
There are many type hints but mypy was not happy so this commit makes it clean again and adds it as part of the GH workflow so that we keep it clean. The registries will have to be redone, they are currently not type safe but we only have a single target right now so it's probably okay for now to just use a very simple approach


supakeen (approved):
Very cool, thank you!

PR #150, opened 2024-06-27
otk: add support for `otk -w duplicate-definition`

mvo5:
We describe in `01-directive.md` that we support warnings via `-w duplicate-definition` - this commit makes this real.

[edit: there is also a typo fix in the warning]


supakeen (approved):
Nice.

PR #152, opened 2024-06-27
otk/command: fix handling of argument '-t'

schuellerf:
...including some tests - which could be written more readable I guess

### DIFF
--- a/test/test_command.py
+++ b/test/test_command.py
@@ -68,84 +68,14 @@ def test_parse_commands_failure(capsys, command, sys_exit_code, sys_exit_message
     assert sys_exit_message in captured_stderr
 
 
-@pytest.mark.parametrize(
-    "data",
-    [
-        # Select no target - only one available
-        {"command": ["compile", "-o", "OUTPUTFILE", "INPUTFILE"],
-         "input_data": """otk.version: "1"
+TEST_ARGUMENT_T_INPUT = """otk.version: "1"
 
 otk.target.osbuild.name:
   pipelines:
     - "test"
-""",
-         "output_data": """{
-  "pipelines": [
-    "test"
-  ],
-  "version": "2",
-  "sources": {}
-}""",
-         "ret_expected": 0,
-         "log_expected": ""
-         },
+"""
 
-        # Select the one available target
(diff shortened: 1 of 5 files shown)


test/test_command.py
```diff
@@ -66,3 +66,110 @@ def test_parse_commands_failure(capsys, command, sys_exit_code, sys_exit_message
...
     captured_stderr = capsys.readouterr().err
     assert sys_exit_message in captured_stderr
+
+
+@pytest.mark.parametrize(
+    "data",
+    [
+        # Select no target - only one available
+        {"command": ["compile", "-o", "OUTPUTFILE", "INPUTFILE"],
+         "input_data": """otk.version: "1"
```
mvo5, line 76: (quick drive-by) I see this 
"""otk.version: "1"

otk.target.osbuild.name:
  pipelines:
    - "test"
"""
a lot - could it be extracted into a var and just used as a var like `minimal_otk` and then (maybe) `minimal_otk + "whatever needs adding" - this would probably make the reading of this easier.

Alternatively there is test/test_examples.py and test/data/error, could some be put there (I guess there is no way to manipulate what `-t` is used there so mabe that won't work).

schuellerf: I thought so too - not sure if this looks better - I'll check again
Second option won't work easily, that's right

supakeen (approved):
Thanks!

test/test_command.py
```diff
@@ -66,3 +66,96 @@ def test_parse_commands_failure(capsys, command, sys_exit_code, sys_exit_message
...
+
+
+TEST_ARGUMENT_T_INPUT = """otk.version: "1"
+
+otk.target.osbuild.name:
+  pipelines:
+    - "test"
+"""
+
+TEST_ARGUMENT_T_INPUT_TWO_TARGETS = """otk.version: "1"
```
mvo5, line 78: Could this now be written as:
```python
TEST_ARGUMENT_T_INPUT_TWO_TARGETS = TEST_ARGUMENT_T_INPUT  + """
otk.target.osbuild.name1:
  pipelines:
    - "test1"
 """"
 ?
```

mvo5: (obviously a nitpick but I really like to avoid g duplication (within reason) :)

PR #154, opened 2024-06-27
otk: use typing.Optional instead of Union[,None]

mvo5:
This commit switches the the shorter and easier (IMHO) to read typing.Optional[] type and uses that instead of typing.Union[]. The are equivalent otherwise.


supakeen (approved):
Right yea, this was due to using the Union operator before for types.

PR #155, opened 2024-06-27
otk: add more mypy strictness

mvo5:
Add more `mypy` strictness, especially `disallow_incomplete_defs`
is useful. This follows roughly the suggestions from
https://mypy.readthedocs.io/en/stable/existing_code.html#introduce-stricter-options

The next steps would probably be:
```
check_untyped_defs = true
disallow_untyped_calls = True
disallow_untyped_defs = True
```
but `disallow_untyped_defs` would require to be able to exclude
all the file in `test/` for this specific check as it seems a
bit overkill.

[based on https://github.com/osbuild/otk/pull/154 to avoid conflicts]


supakeen (approved):
I like more types :)

PR #156, opened 2024-06-28
otk/command: fix includes with stdin

schuellerf:
I think `stdin` is nice for playing around with snippets and no one would probably mix `stdin` with `otk.includes`
but still, they should work and then include from the current working directory

The implementation is a little messy due to the problem with the duplicate resolve @mvo5 hinted me to.

Maybe we keep this until we have time to cleanup `command.py` and remove calling both `resolve()` and `process_include()`

### DIFF
--- a/src/otk/command.py
+++ b/src/otk/command.py
@@ -42,15 +42,19 @@ def _process(arguments: argparse.Namespace, dry_run: bool) -> int:
         dst = sys.stdout if arguments.output is None else open(arguments.output, "w", encoding="utf-8")
 
     if arguments.input is None:
-        path = pathlib.Path(f"/proc/self/fd/{sys.stdin.fileno()}")
+        input_file = pathlib.Path(f"/proc/self/fd/{sys.stdin.fileno()}")
+        # non-existing fake file to get the base paths for includes right
+        path = pathlib.Path.cwd() / "<stdin>"
+        state = State(path=path)
     else:
-        path = pathlib.Path(arguments.input)
+        input_file = pathlib.Path(arguments.input)
+        path = input_file
+        state = State()
 
     ddw = any(arg in getattr(arguments, "warn", [])
               for arg in ["duplicate-definition", "all"])
     ctx = CommonContext(duplicate_definitions_warning=ddw)
-    state = State()
-    doc = Omnifest(process_include(ctx, state, path))
+    doc = Omnifest(process_include(ctx, state, input_file))
 
     target_available = doc.targets
     target_requested = arguments.target
@@ -83,16 +87,14 @@ def _process(arguments: argparse.Namespace, dry_run: bool) -> int:
     if kind != "osbuild":
         raise ValueError("only target osbuild supported right now")
     spec = OSBuildContext(ctx)
-    target = OSBuildTarget()
-    state = State(path=path)
-    tree = resolve(spec, state, doc.tree[f"{PREFIX_TARGET}{kind}.{name}"])
+    tree = resolve(spec, State(path=path), doc.tree[f"{PREFIX_TARGET}{kind}.{name}"])
 
     if not target.is_valid(tree):
         return 1
 
     # and then output by writing to the output
     if not dry_run:
-        dst.write(target.as_string(spec, tree))
+        dst.write(OSBuildTarget().as_string(spec, tree))
 
     return 0
 
--- a/test/test_compile.py
+++ b/test/test_compile.py
@@ -1,6 +1,6 @@
 import argparse
 import os
-
+import pathlib
 
 from otk.command import compile  # pylint: disable=W0622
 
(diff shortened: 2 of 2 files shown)


src/otk/command.py
```diff
@@ -42,12 +42,16 @@ def _process(arguments: argparse.Namespace, dry_run: bool) -> int:
 
     if arguments.input is None:
         path = pathlib.Path(f"/proc/self/fd/{sys.stdin.fileno()}")
+        # non-existing fake file to get the paths right
+        cwd = pathlib.Path.cwd() / "stdin"
```
mvo5, line 46: (quick-drive-by): maybe `<stdin>`or something here so that it looks less like a real filename (especially in errors that print the path of the fragment)?

src/otk/command.py
```diff
@@ -42,12 +42,16 @@ def _process(arguments: argparse.Namespace, dry_run: bool) -> int:
...
+        state = State(path=cwd)
     else:
         path = pathlib.Path(arguments.input)
+        cwd = path
```
mvo5, line 50: I think using `cwd` is misleading here, in my mind `cwd` stands for "current-working-directory` but here we assign a filename, so /home/mvogt/foo.otk would be `cwd` - this seems to have potential for confusion when causually reading the code.

But of course it's tricky to find a good name because we already have `path` and the fact that we now have/need a second path is confusing. This indicates that something in our abstractions is off, probably worth adding a TODO at least that having two paths to workaround weirdness in the stdin handling is something we should look at again.

src/otk/command.py
```diff
@@ -81,13 +85,11 @@ def _process(arguments: argparse.Namespace, dry_run: bool) -> int:
     if kind != "osbuild":
         raise ValueError("only target osbuild supported right now")
     spec = OSBuildContext(ctx)
-    target = OSBuildTarget()
-    state = State(path=path)
-    tree = resolve(spec, state, doc.tree[f"{PREFIX_TARGET}{kind}.{name}"])
+    tree = resolve(spec, State(path=cwd), doc.tree[f"{PREFIX_TARGET}{kind}.{name}"])
```
mvo5, line 88: Unrelated to this PR - I think this will not work when the `target.osbuild.foo` is in a subdirectoy. I.e. if the manifest looks like this:
```
otk.include "subdir/osbuild.otk"
---
subdir/osbuild.otk:

otk.target.osbuild.foo:
  otk.include "file-in-subdir.yaml"
```
the code will not be able to find `file-in-subdir.yaml` because it points to the current cwd. 

This is already a problem and unrelated to the PR. I think the solution would be to do a single tree resolving and pass the expected target(s) into resove() so that resolve() can just skip over anything otk.target.* that is not part of "target_requested" but this way we would have a complete tree in a single run.

mvo5 (commented):
Thank you for looking into this. TBH I'm no longer a fan of <stdin> for the inputs here because if an otk file with includes compiles or now now depends on if one is int right directory which seems a bit fragile. But if we want to keep it, I guess we need to fix it. I added some ideas/suggestions.

src/otk/command.py
schuellerf: I'm just afraid of '<' in a filename - but yes, it's not really used :-D

src/otk/command.py
schuellerf: In the stdin case it is really only used as current working directory for includes but I'll come up with a better/different name

src/otk/command.py
schuellerf: partly unrelated - so I did a second PR
https://github.com/osbuild/otk/pull/158
but this also works when rebased on this stdin PR 🥳
all good 😏

src/otk/command.py
```diff
@@ -83,16 +87,14 @@ def _process(arguments: argparse.Namespace, dry_run: bool) -> int:
     if kind != "osbuild":
         raise ValueError("only target osbuild supported right now")
     spec = OSBuildContext(ctx)
-    target = OSBuildTarget()
```
supakeen, line 86: `target` is still used on line 92.

src/otk/command.py
mvo5: Heh, yeah, I think it's kinda-the-point to make it scary so that it's clear that it's not a real file if it leaks in e.g. an error message (which I think is quite likely, like `something wong in file "/home/mvogt/otk/stdin` vs `something wong in file "/home/mvogt/otk/<stdin>` (or maybe we can tweak the errors to something like `error in file <stdin> (dir /home/mvogt/otk): foo` i.e. splitting the path then the error woudl be nice for both case :)

src/otk/command.py
mvo5: Sorry, I think I did not express my point very well :( Thanks for writing the test! I added a small tweak the test PR in https://github.com/osbuild/otk/pull/158 and it is now failing  and then I looked a bit more in https://github.com/osbuild/otk/pull/159 that also includes a fix now (not 100% happy yet with #159 but I hope/think it demonstrates the idea(s))

supakeen:
Hi, is this PR still relevant or did we fix it elsewhere? It seems includes work with stdin nowadays.

schuellerf:
> Hi, is this PR still relevant or did we fix it elsewhere? It seems includes work with stdin nowadays.

stdin works, includes with stdin (where `pwd` would be correct) doesn't work.
Not sure if we should make it work, or just improve the error message

```
# otk compile < 13-include-in-subdir.yaml
…
otk.error.IncludeNotFoundError: /proc/self/fd/0: file 31535/fd/13-include-in-subdir/target.yaml was not found
```

but I'll close this PR for now and I can fix this behavior if needed.

PR #157, opened 2024-06-28
Implement error for invalid target

schuellerf:
checking that target
- is a dict
- contains 'pipelines'
- does not contain 'version'

checking source is unclear to me - so it's a comment for now

### DIFF
--- a/test/test_example.py
+++ b/test/test_example.py
@@ -31,3 +31,16 @@ def test_errors(src_yaml):
     with pytest.raises(Exception) as exception:
         command.compile(ns)
     assert expected in str(exception.value)
+
+
+@pytest.mark.parametrize("src_yaml",
+                         [str(path) for path in (pathlib.Path(__file__).parent / "data/log_fatal").glob("*.yaml")])
+def test_log_fatal(caplog, src_yaml):
+    src_yaml = pathlib.Path(src_yaml)
+    expected = src_yaml.with_suffix(".err").read_text(encoding="utf8").strip()
+    ns = argparse.Namespace(input=src_yaml, output="/dev/null", target="osbuild.name")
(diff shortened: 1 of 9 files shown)


supakeen (changes requested):
From these checks only "being a `dict`" and having a `version` is absolutely required by `osbuild`. Let's keep our checks to that as well?

supakeen (approved):
ty

test/test_example.py
```diff
@@ -31,3 +31,16 @@ def test_errors(src_yaml):
     with pytest.raises(Exception) as exception:
         command.compile(ns)
     assert expected in str(exception.value)
+
+
+@pytest.mark.parametrize("src_yaml",
+                         [str(path) for path in (pathlib.Path(__file__).parent / "data/log_fatal").glob("*.yaml")])
```
mvo5, line 37: Silly question, why do we have a `test/data/error` and `test/data/log_fatal` dir? What is the difference, why do we need two?

schuellerf: at the most upper call there is no exception - only ret == 1 and a fatal log

mvo5: Why would we do a different user experience, why would we use different code patterns? Are the errors actually different to justify this or is there a different reason? IMHO consistency in both code and user experience is important so either we use the same pattern(s) or if we don't we should at least have a code comment explaining why they diverge. (sorry I hope this not come accross too nitpicky/weird)

mvo5: Sorry, I may have been a bit terse and I see we already mixing some validation that raises exceptions and some that writes log messages  - I guess what I try to get to is this is maybe a good opportunity to step back and see if we can unify the error reporting to use the same pattern(s) :)

PR #158, opened 2024-06-28
test/data/base: add test for relative subdir include

schuellerf:
adding a test for an include relative to a sub directory
seems to work fine


mvo5:
> adding a test for an include relative to a sub directory seems to work fine

I pushed a small tweak, the key for this that the "target" definition must be inside the subdir, all is fine from the top-level.

mvo5:
This should be fixed with #161 - if/once that is merged this one can be closed (the failing test is part of the PR that fixes the issue).

supakeen:
#161 is in the queue.

PR #161, opened 2024-06-28
otk: resolve targets in a full pass

mvo5:
This commit changes the code so that the tree for a specific target
is resolved in a full pass. This fixes the issue that relative includes
in included targets will fail (e.g. test/data/base/13-include-in-subdir).

The commit also includes some discussion points and XXX/questions
that can be discussed/resolved later.

Followup for https://github.com/osbuild/otk/pull/158 

[edit: this is ready now I think, I extracted some code into document.py, I'm not 100% sure it's the best approach but it feels like a nice step forward as it solves some duplication and moves some of the low-level mechanics into a single place inside document.py but I'm definitely open for idas]


supakeen (dismissed):
I am not entirely happy about the way we're doing the 'double resolve' both through the Omnifest object but I also don't have any better plans at the moment, let's move forward :)

src/otk/command.py
```diff
@@ -46,11 +41,14 @@ def _process(arguments: argparse.Namespace, dry_run: bool) -> int:
...
-              for arg in ["duplicate-definition", "all"])
-    ctx = CommonContext(duplicate_definitions_warning=ddw)
-    state = State()
-    doc = Omnifest(process_include(ctx, state, path))
+    warn_duplicated_defs = any(arg in getattr(arguments, "warn", [])
+                               for arg in ["duplicate-definition", "all"])
+    # First pass of resolving the otk file is "shallow", it will not run
+    # externals and not resolve anything under otk.target.*
+    #
+    # It only exists as convinience for the user so that they do not need
```
achilleas-k, line 49:
```suggestion
    # It only exists as convenience for the user so that they do not need
```

achilleas-k (approved):
LGTM.  Small typo in comment though :)

Looking at all this, I'm thinking we're still not at the perfect solution with respect to target selection and one- vs two-pass behaviour.  I think there's some interesting valid cases that we can't cover with the dry-run as first pass.  But we can discuss all of this separately and with proper examples.

Thanks!

src/otk/command.py
mvo5: Thanks! Fixed and (force) pushed

supakeen (approved):

PR #162, opened 2024-07-01
otk,test: support otk.external inside otk.define

supakeen:
Adds a test that uses a mocked mirror external that returns the subtree directly inside an `otk.define` block. The variable is not available while it should be.


supakeen:
Looks amazing and works; however I can't ack my own PR so can you do that by proxy @mvo5?

mvo5 (approved):
Thank you

PR #163, opened 2024-07-01
otk: error for top-level keys without targets

mvo5:
This commit disallows top level keys that have no targets and are not `otk.` specific. I.e. it disallows something like:
```yaml
otk.version: 1
otk.target.osbuild:

random:
 key: value
```

The rational is that a construct like this may confuse users as the `random:` subtree will not appear in any output or have any other effect.

### DIFF
--- a/src/otk/document.py
+++ b/src/otk/document.py
@@ -2,8 +2,8 @@ import logging
 from copy import deepcopy
 from typing import Any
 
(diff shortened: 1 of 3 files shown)


supakeen (approved):
Agreed.

supakeen (approved):

PR #164, opened 2024-07-01
test: simplify naming of yaml/targets

mvo5:
We allow the use of `otk.define.foo` or `otk.target.osbuild.foo`
to ensure when multiple `otk.defines` are done in a single file
it is still valid yaml. However in a lot of our examples we only
have a single `otk.define` in the file or only a single target.
Here there is no need to append a postfix.

This commit simplifies these occurances and removes unneeded
postfixes. They are still valid and some tests needs them but
the are no longer there by default.

We should probably do the same for the `examples/**`.

### DIFF
--- a/test/data/base/00-empty.yaml
+++ b/test/data/base/00-empty.yaml
@@ -5,6 +5,6 @@
 # The results of compiling this omnifest are available in `00-empty.json`.
 otk.version: "1"
 
-otk.target.osbuild.name:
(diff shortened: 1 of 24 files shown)


supakeen:
Can be undrafted/rebased. #161 has landed.

supakeen (approved):
Thank you, much more clear.

PR #165, opened 2024-07-03
test: add test that has two includes and fix the resulting fallout

mvo5:
This commit fixes the bug that the loop over a dict returns too early for `otk.target` and for `otk.include` - both need to continue iterating over the dict.

As fallout from this the command.py code also needs a tweak and "duplicated_defintion" warnings are now only printed in the final resolve run (not in the target-discovery phase).

A regression test is added based on the excellent example from Simon (thank you!).


supakeen (approved):
Nice, thank you.

PR #166, opened 2024-07-04
test/data: different types for external defines

supakeen:
Rename some keys and add:

- a list-type being returned by an external.
- a nested-dict being returned by an external.


schuellerf (approved):
LGTM

PR #167, opened 2024-07-04
fedora: fix fedora compile

mgold1234:
this commit fix compile problems in minimal-40-aarch64.yaml and minimal-40-x86_64.yaml files.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -28,3 +28,7 @@ jobs:
           pip install .[dev]
       - name: Run unit tests
         run: PYTHONPATH=./src pytest
+      - name: Validate OTK configuration
+        run: |
+          otk -v validate example/fedora/minimal-40-aarch64.yaml
+          otk -v validate example/fedora/minimal-40-x86_64.yaml
--- a/example/fedora/minimal-40-aarch64.yaml
+++ b/example/fedora/minimal-40-aarch64.yaml
@@ -7,24 +7,16 @@ otk.define:
 
   packages:
     # Repositories to fetch packages from
-    repositories:
-      otk.include: "repository/${version}/repositories.yaml"
     # GPG keys to verify packages with
-    keys:
-      otk.include: "repository/${version}/keys.yaml"
     # These packages are used in the buildroot
     buildroot:
       docs: false
       weak: false
-      packages:
-        otk.include: "packages/${version}/_buildroot.yaml"
     # These packages are used for the operating system tree which is what ends
     # up in the outputs.
     tree:
       docs: false
       weak: false
-      packages:
-        otk.include: "packages/${version}/minimal.yaml"
   filesystem:
     root:
       uuid: "6e4ff95f-f662-45ee-a82a-bdf44a2d0b75"
@@ -36,7 +28,7 @@ otk.define:
       vfs_type: "ext4"
       path: "/boot"
       options: "defaults"
-    boot-efi:
+    boot_efi:
       uuid: "7B77-95E7"
       vfs_type: "vfat"
       path: "/boot/efi"
@@ -44,8 +36,9 @@ otk.define:
       passno: 2
 
 otk.target.osbuild.qcow2:
+  otk.include: "repository/${version}/repositories.yaml"
   pipelines:
-    - otk.include: "osbuild/buildroot.yaml"
-    - otk.include: "osbuild/pipeline/tree.yaml"
+    # otk.include: "osbuild/buildroot.yaml"
+    # otk.include: "osbuild/pipeline/tree.yaml"
     - otk.include: "osbuild/pipeline/raw.yaml"
     - otk.include: "osbuild/pipeline/xz.yaml"
--- a/example/fedora/minimal-40-x86_64.yaml
+++ b/example/fedora/minimal-40-x86_64.yaml
@@ -7,24 +7,16 @@ otk.define:
 
   packages:
     # Repositories to fetch packages from
-    repositories:
-      otk.include: "repository/${version}/repositories.yaml"
     # GPG keys to verify packages with
-    keys:
-      otk.include: "repository/${version}/keys.yaml"
     # These packages are used in the buildroot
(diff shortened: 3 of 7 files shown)


.github/workflows/test.yml
```diff
@@ -28,3 +28,12 @@ jobs:
           pip install .[dev]
       - name: Run unit tests
         run: PYTHONPATH=./src pytest
+      - name: Validate OTK configuration
+        run: otk compile example/fedora/minimal-40-x86_64.yaml
```
mvo5, line 32: Maybe use: `run: |` and do an otk validate for both the minimal-40-x86` and the `-aarch64` one? We *could* (but don't have to) use `otk -v validate ...` to get slightly more debug output if something fails.

example/fedora/minimal-40-aarch64.yaml
```diff
@@ -45,7 +37,6 @@ otk.define:
 
 otk.target.osbuild.qcow2:
   pipelines:
-    - otk.include: "osbuild/buildroot.yaml"
-    - otk.include: "osbuild/pipeline/tree.yaml"
+    - otk.include: "repository/${version}/repositories.yaml"
```
mvo5, line 40: I personally would move this otk.include before the pipeline, it does not matter for the output but feels more "correct" because the repository defines are not really part of a pipeline, more some setup to enable the pipelines to work (but people can have different opinions here I suppose).

example/fedora/minimal-40-x86_64.yaml
```diff
@@ -45,7 +37,6 @@ otk.define:
 
 otk.target.osbuild.qcow2:
   pipelines:
-    - otk.include: "osbuild/buildroot.yaml"
-    - otk.include: "osbuild/pipeline/tree.yaml"
+    - otk.include: "repository/${version}/repositories.yaml"
```
mvo5, line 40: (some comment as above)

example/fedora/repository/40/repositories.yaml
```diff
@@ -1,2 +1,5 @@
-- id: "fedora"
-  metalink: https://mirrors.fedoraproject.org/metalink?repo=fedora-${version}&arch=${architecture}
+otk.define:
++  packages:
```
mvo5, line 2: it looks like there are some copy/paste leftover `+` here at the start of the line?

.github/workflows/test.yml
```diff
@@ -28,3 +28,7 @@ jobs:
           pip install .[dev]
       - name: Run unit tests
         run: PYTHONPATH=./src pytest
+      - name: Validate OTK configuration
+        run: |
+          otk -v validate example/fedora/minimal-40-aarch64.yaml && \
```
mvo5, line 33: (nitpick) in a `run: |` like this the `\` and `&&` are not needed, this could be:
```sh
          otk -v validate example/fedora/minimal-40-aarch64.yaml
          otk -v validate example/fedora/minimal-40-x86_64.yaml
```

example/fedora/minimal-40-aarch64.yaml
```diff
@@ -36,16 +28,15 @@ otk.define:
...
       vfs_type: "vfat"
       path: "/boot/efi"
       options: "defaults,uid=0,gid=0,umask=077,shortname=winnt"
       passno: 2
 
 otk.target.osbuild.qcow2:
+  otk.include: "repository/${version}/repositories.yaml"
   pipelines:
-    - otk.include: "osbuild/buildroot.yaml"
-    - otk.include: "osbuild/pipeline/tree.yaml"
```
mvo5, line 49: Thanks for moving the `  otk.include: "repository/${version}/repositories.yaml"` out -  the other two inludes (`buildroot.yaml` and `tree.yaml`) we want to keep

mvo5 (commented):
Thank you! This looks nice! Two small tweaks and it's good to go I think :)

example/fedora/minimal-40-aarch64.yaml
mgold1234: with those rows I got this error:
  File "/Users/michalgold/git/edge-management/image-builder-project/otk/src/otk/transform.py", line 39, in resolve
    return resolve_list(ctx, state, data)
  File "/Users/michalgold/git/edge-management/image-builder-project/otk/src/otk/transform.py", line 129, in resolve_list
    return [resolve(ctx, state, val) for val in tree]
  File "/Users/michalgold/git/edge-management/image-builder-project/otk/src/otk/transform.py", line 129, in <listcomp>
    return [resolve(ctx, state, val) for val in tree]
  File "/Users/michalgold/git/edge-management/image-builder-project/otk/src/otk/transform.py", line 37, in resolve
    return resolve_dict(ctx, state, data)
  File "/Users/michalgold/git/edge-management/image-builder-project/otk/src/otk/transform.py", line 116, in resolve_dict
    return resolve(ctx, state, call(key, resolve(ctx, state, val)))
  File "/Users/michalgold/git/edge-management/image-builder-project/otk/src/otk/transform.py", line 37, in resolve
    return resolve_dict(ctx, state, data)
  File "/Users/michalgold/git/edge-management/image-builder-project/otk/src/otk/transform.py", line 67, in resolve_dict
  File "/Users/michalgold/git/edge-management/image-builder-project/otk/src/otk/tree.py", line 23, in wrapper
    return function(*args, **kwargs)
  File "/Users/michalgold/git/edge-management/image-builder-project/otk/src/otk/transform.py", line 285, in substitute_vars
    return ctx.variable(m.group("name"))
  File "/Users/michalgold/git/edge-management/image-builder-project/otk/src/otk/context.py", line 103, in variable
    raise TransformVariableLookupError(f"could not find {parts!r}")
otk.error.TransformVariableLookupError: could not find ['packages', 'keys']

mvo5: I see - thank you! Maybe just comment them out in this case? This way we will not forget about them :)

mvo5 (approved):
Thank you!

PR #168, opened 2024-07-04
test: add tests for directives that do not allow siblings

mvo5:
It is not allowed to have multiple `otk.op` or `otk.external` in the same dict today, add a test that makes this explicit.

Fwiw, it's not clear to me why we disallow this but given that it seems a concious decision let's ensure we not change it by accident.


achilleas-k:
> Fwiw, it's not clear to me why we disallow this but given that it seems a concious decision let's ensure we not change it by accident.

I share your concern/scepticism.  Great to have a test for it though so if we ever change it it's clearly tested and we can avoid accidental side effects :heart:

achilleas-k (approved):

PR #169, opened 2024-07-05
example/centos: fix it

supakeen:
Draft PR with my work to make the CentOS targets work excluding the pending disk externals.

Based on: #170.


supakeen:
Going to reimplement this in smaller bits/commits.

PR #174, opened 2024-07-08
otk: allow `otk.define={"a": {}}`

mvo5:
This commit fixes the special case that an define can be an empty
dict.

Base on https://github.com/osbuild/otk/pull/173

### DIFF
--- a/src/otk/transform.py
+++ b/src/otk/transform.py
@@ -153,6 +153,9 @@ def process_defines(ctx: Context, state: State, tree: Any) -> None:
     if tree is None:
         log.warning("empty otk.define in %s", state.path)
         return
+    if tree == {}:
+        ctx.define(state.define_subkey(), {})
+        return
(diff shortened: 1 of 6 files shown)


achilleas-k (approved):

src/otk/transform.py
```diff
@@ -153,6 +153,9 @@ def process_defines(ctx: Context, state: State, tree: Any) -> None:
     if tree is None:
         log.warning("empty otk.define in %s", state.path)
         return
+    if tree == {}:
```
supakeen, line 156: Do we want to do this for `[]` as well?

achilleas-k: I think we should.

achilleas-k: But wait, here `tree` is always a dictionary.  I think `[]` might work as is right now.

PR #178, opened 2024-07-30
pre-commit: add pylint to be checked

schuellerf:
`pylint` is part of github actions, so it should be checked locally to avoid problems


supakeen (commented):
As this moves under `pre-commit` can we remove the separate call in the `Makefile` under the `lint` target or we're doing duplicate work :)

PR #180, opened 2024-08-01
Makefile: add hint if pre-commit is not installed

schuellerf:
Do you think this is too much?
or just leave it up to the users to checkout the README.md


supakeen (approved):
It's not too much.

PR #181, opened 2024-08-01
pre-commit: add mypy

schuellerf:
…while replacing the call also in `Makefile`


supakeen:
I'll cc @mvo5 on this as we are the `mypy` fans. I think we wanted to keep `mypy` optional so not adding it as a required check (this will make it run in `ci` as well right?) is probably better.

supakeen (approved):
Ah no, disregard the previous comment we *do* run it in CI so I'm positive on including it in `pre-commit` as it's fast and a required check.

PR #185, opened 2024-09-02
src: add `--libdir` to override buildin libdir

mvo5:
src: add `--libdir` to override buildin libdir

This allows the local developer to easily run local externals. Similar
to the experience with osbuild and given that otk developers will
most likely also run osbuild having the two similar is probably
a good thing.

[draft as the this is only very lightly tested (and not lint happy)]

### DIFF
--- a/src/otk/command.py
+++ b/src/otk/command.py
@@ -1,5 +1,6 @@
 import argparse
 import logging
+import os
 import pathlib
 import sys
 from typing import List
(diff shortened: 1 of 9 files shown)


supakeen (commented):
Thank you, my only concern here is adding the local developer case by default to the search path.

supakeen:
I'd support having a relative path but instead of magically inserting them have them behind a single argument flag like `-d` on the main command for `dev`?

mvo5:
> I'd support having a relative path but instead of magically inserting them have them behind a single argument flag like `-d` on the main command for `dev`?

I pushed a new version that adds `--libdir` in the spirit of osbuild, given that the two will often be used in parallel it's probably a good idea to keep them similar.

supakeen:
The type checker needs appeasement then we can get it in. `--libdir` is fine to me. There's no real reason to have `_libdir` + `libdir()` property as we aren't doing anything to it right but that's so minor.

mvo5:
Closing for now as there is a different workaround (using the env)

PR #186, opened 2024-09-02
[RFC] test: test against reference images lib generated  manifest

mvo5:
This is an idea how to make development of the otk manifest
easier to iterate on. The idea is that we filter out the
sources/rpms (we should consider later to keep the rpm names but
not the exact versions) and then compare the stages. This way
we can iterate on the development of the centos-9 example and
always be confident that the reference is matching.

The reference was generated with the `images` library.

The next step would be to do the partition table with this, then
split out stages etc.


achilleas-k (commented):
LGTM do far.  We could probably automate the generation and comparison of the baselines, but perhaps that's for later.

supakeen:
Super nice, can't wait to see where this goes.

test/data/images-ref/centos-9-x86_64-qcow2.ref.yaml
```diff
@@ -0,0 +1,221 @@
...
+        options:
+          file_contexts: etc/selinux/targeted/contexts/files/file_contexts
+          labels:
+            /usr/bin/cp: system_u:object_r:install_exec_t:s0
+  - name: os
+    build: name:build
+    stages:
+      - type: org.osbuild.kernel-cmdline
+        options:
+          root_fs_uuid: d8bb61b8-81cf-4c85-937b-69439a23dc5e
```
schuellerf, line 22: are we already that far with `otk` to generate this number in a variable (dynamically) and use it below?

…mainly to demonstrate this functionality - this might be too far from the current issue, so you can as well ignore this comment

example/centos/centos-9-x86_64-ami.yaml
```diff
@@ -1,37 +1,136 @@
...
-otk.include.centos-9-x86_64: "common/centos-9.yaml"
+otk.target.osbuild:
+  pipelines:
+    - otk.include: "pipeline/build.yaml"
+    - name: os
+      build: name:build
+      stages:
+        - otk.include: "fragment/kernel-cmdline.yaml"
+        - type: org.osbuild.rpm
+        - type: org.osbuild.fix-bls
```
mvo5, line 18: extracting: org.osbuild.fix-bls,org.osbuild.keymap, org.osbuild.timezone would work here but because it's in a list it will look quite awkward

example/centos/centos-9-x86_64-ami.yaml
```diff
@@ -1,37 +1,139 @@
...
+              - nm-cloud-setup.service
+              - nm-cloud-setup.timer
+              - cloud-init
+              - cloud-init-local
+              - cloud-config
+              - cloud-final
+              - reboot.target
+              - tuned
+            default_target: multi-user.target
+        - type: org.osbuild.selinux
```
mvo5, line 134: this should be split into a fragement as it's use with the same options everywhere

example/centos/centos-9-x86_64-ami.yaml
supakeen: It won't be *that* awkward, I believe we already use `otk.op.join` in some places and I used it 'everywhere' in my branch just so fragments could define one or more stages :)

mvo5:
All the relevant parts of this PR got extracted into smaller PRs with proper commits so this one can be closed.

PR #187, opened 2024-09-02
target: don't write `sources` by default

supakeen:
Sources are no longer implicit to the osbuild context but will instead be generated explicitly by externals. Remove them from the target, context, and test cases.

I'm splitting out the work of fixing up CentOS started in #169 in separate PRs so #186 can build upon them without it being an unreviewable mess.


supakeen:
Force pushed after the cherry-pick to have the newer test data files also drop the expected sources key.

mvo5 (approved):
Thank you!

PR #188, opened 2024-09-04
examples,tests: add new `test_against_images_ref.py` test

mvo5:
This PR adds a new test that uses the `images` library to generate reference manifests and then compares the output of an otk run against the reference image. This allows us to refactor the `otk` images and always be confident that the result matches what "images" produces.

It also converts all the examples to be just "verbatim" versions of the images library result so that it can be a useful starting point to refactor them.

Note that currently the reference manifests are part of the repository for convenience. It would be nicer to generate them on the fly but that is hard(er), see the comments in the code.

Note also that we most likely will have to do something about the "{boot,root}_fs_uuid" in the reference images as they will not match with the ones generated by `otk-gen-partition-table` as the calls into `rand` are in a different order yielding predictable but different result between the two. We can cross this bridge when we reach it though and it's fairly simple (or exclude at the diff level).

When we add the `dnfjson` external we will need to teach it to generate fake urls/shas just like `gen-manifests` so that we can compare the results of a real run.

The next steps should be to extract the common pieces into a nice structure (i.e. build https://github.com/osbuild/otk/pull/169 on top of this one) and integrate the externals (i.e. start with partition, c.f. https://github.com/osbuild/otk/pull/186)

[clean split out from https://github.com/osbuild/otk/pull/186]


supakeen (approved):
<3 seems like some of the manifest names don't match what's expected in the testcase but that should be a minor fix.

mvo5:
> <3 seems like some of the manifest names don't match what's expected in the testcase but that should be a minor fix.

Thank you! Yeah, it also looks like "yaml.dump()" is not generating yamllint clean yaml :(

supakeen (approved):

PR #190, opened 2024-09-09
Makefile: add "make external" to have a quick way to build externals

mvo5:
This commit adds a new "make external" command that should help iterate faster by making the images otk externals available easily for development.

[split out from https://github.com/osbuild/otk/pull/185 to unblock the external development]


supakeen (approved):
💯

PR #191, opened 2024-09-09
test: use `sort_keys=False` in the `gen-image-def` helper

mvo5:
When generating the "images" library reference yaml we used the
default yaml.dump() implementation. This will reoder the keys
which is undesirable for us as it moves e.g. the `type: ...`
key at the end of a block. But in a osbuild manifest this is
the one of the most important pieces of information.

So this commit changes the code to not sort the yaml keys.

---

test: re-generate the reference images with yaml
 `sort_keys=False`

Trivial update to use the previous commit:
```
./gen-image-def ~/devel/osbuild/images centos 9 x86_64 ami
./gen-image-def ~/devel/osbuild/images fedora 40 x86_64 minimal_raw
./gen-image-def ~/devel/osbuild/images fedora 40 x86_64 minimal-raw
./gen-image-def ~/devel/osbuild/images fedora 40 aarch64 minimal-raw
```

### DIFF
--- a/test/data/images-ref/centos/9/x86_64/ami/centos_9-x86_64-ami-empty.yaml
+++ b/test/data/images-ref/centos/9/x86_64/ami/centos_9-x86_64-ami-empty.yaml
@@ -1,9 +1,12 @@
+version: '2'
 pipelines:
   - name: build
     runner: org.osbuild.centos9
     stages:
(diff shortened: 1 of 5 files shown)


supakeen (approved):
Looks much more readable, thanks.

PR #193, opened 2024-09-09
example: port `centos-9-x86_64-qcow2.yaml` to use partition  externals

mvo5:
workflow: ensure "make test" builds externals and use in
 workflow

This commit ensures that externals are build during the tests and
that the GH action will test with externals too.

This is slightly hacky as it requires the network. Eventually we
can pull in externals via build-deps or vendor them but for now
this unblocks the otk work.

---

example: port `centos-9-x86_64-qcow2.yaml` to use partition
 externals

This commit uses the various externals to generate partition tables
in the qcow2 image.

---

test: use `OTK_EXTERNAL_PATH` in test_images_ref

Use the locally donwloaded/build `external` in this test to ensure
that the otk acually finds externals.

--- 

test: workaround different RNG results in `gen-manifests`

This is a hacky commit - the OSBUILD_TESTING_RNG_SEED is set for both
"gen-manifests" and "otk-gen-partition-tables" but the partition
table code is called slightly differently which means that the UUIDs
get out of sync. We need this (f)ugly mapping here to fix that.

Fortunately both sides are deterministic so this mapping is hopefully
enough.

### DIFF
--- a/example/centos/centos-9-x86_64-qcow2.yaml
+++ b/example/centos/centos-9-x86_64-qcow2.yaml
@@ -1,5 +1,11 @@
 otk.version: "1"
 
+otk.define:
+  filesystem:
+    modifications:
+    # empty
+otk.include: "common/gen-partition-table-x86_64.yaml"
+
 otk.target.osbuild:
   pipelines:
     - name: build
@@ -99,7 +105,7 @@ otk.target.osbuild:
       stages:
         - type: org.osbuild.kernel-cmdline
           options:
-            root_fs_uuid: dde1466f-f75b-4f05-bcf6-8386564c6f79
+            root_fs_uuid: ${filesystem.const.partition_map.root.uuid}
             kernel_opts: console=tty0 console=ttyS0,115200n8 no_timer_check net.ifnames=0
         - type: org.osbuild.rpm
           inputs:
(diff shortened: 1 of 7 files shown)


example/centos/centos-9-x86_64-qcow2.yaml
```diff
@@ -251,130 +243,39 @@ otk.target.osbuild:
...
+                      to: mount://-/
+                devices:
+                  ${fs_options.devices}
+                mounts:
+                  ${fs_options.mounts}
+              - type: org.osbuild.grub2.inst
+                options:
+                  filename: disk.img
+                  platform: i386-pc
+                  location: 2048
```
mvo5, line 269: These will also have to be somehow defined via otk-gen-partition I think

supakeen: We can probably look into the usability/niceness of it better after we have some working examples so I wouldn't block this PR on this :)

mvo5: Agreed, I opened https://github.com/osbuild/images/pull/916 but the change can be a folloup.

supakeen (approved):

PR #196, opened 2024-09-12
rewrite depsolve external

supakeen:
This PR splits out the depsolve external along the same idiom that the partition externals follow. Draft because it needs a manifest as well.


supakeen:
Note: mock GPG keys because they'll move into the manifest defines.

src/otk_external_osbuild/command/depsolve_dnf4.py
```diff
@@ -0,0 +1,66 @@
+"""`otk-osbuild` is the [otk](https://github.com/osbuild/otk) external to work
+with [osbuild](https://github.com/osbuild/osbuild) manifests.
+
+THIS FILE IS VERY, VERY PROOF OF CONCEPT AND NEEDS TO BE CLEANED UP"""
```
mvo5, line 4: Just as high level feedback, for the partition stage we called the thing that generates the otk.defines `gen-partition` and the thing that makes the stages `make-` here it seems to be different, I think we should be consistent.

The gen/make is not super great, might be worth brainstorm/bikeshed a bit about it, some alternative ideas:
- `define-*` / `make-*`
- `fill-` / `make-`
- `fill-` / `do-`
- `construct-` / `make-` (but a bit long)

Maybe someone like @achilleas-k has more ideas but I would prefer if we start consistent and then change everywhere at the same time.

supakeen: Yea I assume we'll bikeshed about the naming; I created a [PR in images](https://github.com/osbuild/images/pull/922) as well so we can discuss there or here.

mvo5: Maybe worth starting with the existing schema then once the new externals are integrated we have a more holistic view of how the rename fits into the otk files and we postpone the bikeshed (also to avoid yet another rename if we change our minds later)

supakeen: The existing being the `otk-` naming with `make` and `gen` from the Go externals?

supakeen:
Needs:
- [ ] Tests to be able to call the Python externals as well
- [ ] Updated comments in the externals.
- [ ] Maybe schemas, or followup?

mvo5 (commented):
Thank you! It's super nice to see this coming together. A few suggestions/ideas inline for your consideration.

mvo5:
> Needs:
> 
>     * [ ]  Tests to be able to call the Python externals as well
> 
>     * [ ]  Updated comments in the externals.
> 
>     * [ ]  Maybe schemas, or followup?

Fwiw, I think the schema is fine in a followup as e.g. images or osbuild do not use dnfjson schemas either.

example/centos/centos-9-x86_64-ami.yaml
```diff
@@ -6,98 +6,132 @@ otk.define:
...
...kYR3IRZLXlWqeOi/+ezYOW0m\nvufrkcvsh+TKlVVnuwmEPjJ8mwUSpsLdfPJo1DHsd8FS03SCKPaXFdD7ePfEjiYk\nnHpQaKE01aWVSLUiygn7F7rYemGqV9Vt7tBw5pz0vqSC72a5E3zFzIIuHx6aANry\nGat3aqU3qtBXOrA/dPkX9cWE+UR5wo/A2UdKJZLlGhM2WRJ3ltmGT48V9CeS6N9Y\nm4CKdzvg7EWjlTlFrd/8WJ2KoqOE9leDPeXRPncubJfJ6LLIHyG09h9kKQARAQAB\ntDpDZW50T1MgKENlbnRPUyBPZmZpY2lhbCBTaWduaW5nIEtleSkgPHNlY3VyaXR5\nQGNlbnRvcy5vcmc+iQI3BBMBAgAhBQJczFsZAhsDBgsJCAcDAgYVCAIJCgsDFgIB\nAh4BAheAAAoJEAW1VbOEg8ZdjOsP/2ygSxH9jqffOU9SKyJDlraL2gIutqZ3B8pl\nGy/Qnb9QD1EJVb4ZxOEhcY2W9VJfIpnf3yBuAto7zvKe/G1nxH4Bt6WTJQCkUjcs\nN3qPWsx1VslsAEz7bXGiHym6Ay4xF28bQ9XYIokIQXd0T2rD3/lNGxNtORZ2bKjD\nvOzYzvh2idUIY1DgGWJ11gtHFIA9CvHcW+SMPEhkcKZJAO51ayFBqTSSpiorVwTq\na0cB+cgmCQOI4/MY+kIvzoexfG7xhkUqe0wxmph9RQQxlTbNQDCdaxSgwbF2T+gw\nbyaDvkS4xtR6Soj7BKjKAmcnf5fn4C5Or0KLUqMzBtDMbfQQihn62iZJN6ZZ/4dg\nq4HTqyVpyuzMXsFpJ9L/FqH2DJ4exGGpBv00ba/Zauy7GsqOc5PnNBsYaHCply0X\n407DRx51t9YwYI/ttValuehq9+gRJpOTTKp6AjZn/a5Yt3h6jDgpNfM/EyLFIY9z\nV6CXqQQ/8JRvaik/JsGCf+eeLZOw4koIjZGEAg04iuyNTjhx0e/QHEVcYAqNLhXG\nrCTTbCn3NSUO9qxEXC+K/1m1kaXoCGA0UWlVGZ1JSifbbMx0yxq/brpEZPUYm+32\no8XfbocBWljFUJ+6aljTvZ3LQLKTSPW7TFO+GXycAOmCGhlXh2tlc6iTc41PACqy\nyy+mHmSv\n=kkH7\n-----END PGP PUBLIC KEY BLOCK-----"
```
mvo5, line 133: One quick thought - I wonder if we can `otk.include` the keys given that both ami/qcow2 seems to use the same so that we only have a single copy in the repo?

example/centos/centos-9-x86_64-ami.yaml
```diff
@@ -6,98 +6,132 @@ otk.define:
...
+        module_platform_id: f40
+        releasever: "40"
+        repositories:
+          - id: "AppStream"
+            baseurl: "https://rpmrepo.osbuild.org/v2/mirror/public/el9/cs9-x86_64-appstream-20240901"
+          - id: "BaseOS"
+            baseurl: "https://rpmrepo.osbuild.org/v2/mirror/public/el9/cs9-x86_64-baseos-20240901"
+        packages:
+          include:
+            - selinux-policy-targeted
```
mvo5, line 22: (super nitpick) woudl it make sense to sort the pkgs?

supakeen:
- [ ] Use `.const.internal` keys on the data.
- [ ] Provide `.const.kernel` (or `.kernel`?) containing kernel version and name.
- [ ] Rename to fit current idiom.

supakeen:
I've opted for a very similar approach to `rpmmd.GetVerStrFromPackageSpecList`; we get an additional non-internal map keyed by package names that contains the parts that can be used to form the necessary information.

Perhaps we *only* want the (N)EVRA in there?

example/centos/centos-9-x86_64-qcow2.yaml
```diff
@@ -6,97 +6,131 @@ otk.define:
...
+            - psmisc
+            - python3-jsonschema
+            - qemu-guest-agent
+            - redhat-release
+            - redhat-release-eula
+            - rsync
+            - tar
+            - tuned
+            - tcpdump
+          exclude:
```
mvo5, line 86: Finding a way to test this would be really good, this will probably require changes in the `gen-manifests` mock dnfjson mode in images though so probably followup material but IMHO quite important to ensure we really have the same images.

src/otk_external_osbuild/command/gen_depsolve_dnf4.py
```diff
@@ -0,0 +1,101 @@
...
+    """Transform the output of `osbuild-depsolve-dnf4` to the output format
+    expected of this external."""
+
+    data = {"tree": {"const": {"internal": {}}}}
+
+    # Expose the direct output as internal-only data, this can be used by
+    # other externals.
+    data["tree"]["const"]["internal"]["packages"] = packages
+
+    # We also store all resolved packages and some meta information about
```
mvo5, line 18: This adds a very broad API to this otk external, I personally would start more targeted, we need the kernel name and version and in `images` the kernel is handled specially only. So maybe we could do the same here? Something like just expose a `const.kernel.{name,version}`? We will need to adjust `fragments/grub2.yam` to match the kernel and the string there includes `0-0.noarch` right now which also indicates that we need to tweak our tests to make this better testable.

Here is an alternative idea https://github.com/osbuild/otk/commit/5f378bb2631bc4cdb7cd6ae40d4cf1f0ec38b877

src/otk_external_osbuild/command/gen_depsolve_dnf4.py
```diff
@@ -0,0 +1,101 @@
...
+            "arch": "",
+        }
+        for p in packages
+    ]
+
+
+def root():
+    data = json.loads(sys.stdin.read())
+    tree = data["tree"]
+    mock = "OTK_UNDER_TEST" in os.environ
```
mvo5, line 51: Any downside of writing this slightly compacter into:
```python
    if "OTK_UNDER_TEST" in os.environ:
        # When we are under test we don't call the depsolver at all and instead
        # return a mocked list of things
        packages = mockdata(tree["packages"]["include"])
        ...
``` 
(i.e. `mock` is assigned but never used again, only is this diff, by moving it all into the same line it feels easier to read as there is less context to care about)

mvo5 (commented):
Thanks, this moving along nicely, some comemnts inline.I really want to look a bit more at the gen-manifests reference image generator to get more meaningful mock data, right now we do not test anything like excludes or we don't mock versions or arch which is important for the testing of the grub2 kernel bits.

src/otk_external_osbuild/command/gen_depsolve_dnf4.py
supakeen: No reason to, I'll adjust!

src/otk_external_osbuild/command/gen_depsolve_dnf4.py
supakeen: We can set mock data to always be `0`, `0`, and `noarch` instead of the empty strings they are now.

I chose the broad API to not limit ourselves here also taking some inspiration from `osbuild-mpp` where it is currently possible to get the version of any package.

supakeen:
Addressed the mock thing in a separate commit (can be squashed if we want) but keeps the review a bit cleaner for now. Last outstanding thing is the package versions, do we want all of them to be available of just the provided kernel package name?

src/otk_external_osbuild/command/gen_depsolve_dnf4.py
mvo5: Thanks, that is interesting! How is this used by folks that use `osbuild-mpp` - are there more use-cases then the kernel and if so, which ones? Would love to learn more here.

supakeen:
This PR has been fully split out into #204, #203, #202.

src/otk_external_osbuild/command/gen_depsolve_dnf4.py
achilleas-k: The only call to the equivalent function in images that's not for looking up the kernel is to check if `dracut-config-rescue` exists in the package set, but it doesn't care about version.
https://github.com/osbuild/images/blob/67a2b2f65a2e0969039a3de17bf2c266390d01ab/pkg/manifest/os.go#L651

PR #198, opened 2024-09-16
example/fedora: minimal-raw for x86_64

mvo5:
This commit adds anoter example to the mix of distro examples. Here it is the minimal fedora raw image that is important for the community. It is a also a useful way to explore how much of the osbuild fragments/pipelines can be shared access (rpm-type) distros.


supakeen:
I only 'disagree' with cross-distro sharing for this. Perhaps we should close this PR and revisit now that so much has changed in the CentOS examples?

mvo5:
Closing them for now as there are conlicts and new pattern.

PR #199, opened 2024-09-16
example/fedora: minimal-raw for aarch64

mvo5:
This adds another important example to the otk files. The first that
is not x86_64. It is also important for the community and it allows
us to explore how much fragements/pipelines we can share accross
the images.

It unconverted missing support for creating ESP partitions on  DOS partition tables in the otk-gen-partition-table external. Until this is fixed this needs to be draft.
[edit: this needs https://github.com/osbuild/images/pull/934 and https://github.com/osbuild/otk/pull/198]


supakeen:
I only 'disagree' with cross-distro sharing for this. Perhaps we should close this PR and revisit now that so much has changed in the CentOS examples?

PR #202, opened 2024-09-16
gen_depsolve_dnf4: include version/release/arch for kernel

mvo5:
This is draft with some ideas about the `gen-depsolve-dnf4` handling of the kernel package.

The basic ideas:
- mock more in the gen-manifests so that our mock dnfjson has non-empty version/release too
- handle the kernel special in gen-depsolve-dnf4 as it is handled special in images too and provide release/version/arch to otk as an API
- use this data in the grub2 fragment to have a generic way to reference the kernel

This needs https://github.com/osbuild/images/pull/937 and only the last two commits are relevant, the rest is based on https://github.com/osbuild/otk/pull/196


mvo5:
This is now obsolete with #226

PR #204, opened 2024-09-17
example: centos9-qcow/ami with depsolve

mvo5:
This is another splitout of #196 - now that we have the resolver we use it in the otk files.

[draft as I think we want to de-duplicate and sort the packages list]

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,6 +21,9 @@ format:
 
 .PHONY: test
 test: external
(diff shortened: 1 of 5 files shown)


supakeen:
Undrafted as I've sorted and deduplicated the package sets here.

supakeen (approved):

PR #207, opened 2024-09-18
many: test for `exclude` when comparing images/otk manifests

mvo5:
This PR uses https://github.com/osbuild/images/pull/938 so compare that the dnfjson excludes are also identical between the osbuild manifests that are generated by images and by otk. 

It also contain a fix/update for the images pr#943, pr#937


supakeen (approved):
Very cool :)

PR #206, opened 2024-09-17
centos draft (COMPOSER-2295)

supakeen:

PR #208, opened 2024-09-18
make: include new externals

supakeen:
Include the recently merged container resolver and ostree commit resolver externals.

Necessary for #206, was implemented in: https://github.com/osbuild/images/pull/936 and https://github.com/osbuild/images/pull/935


bcl (approved):
Looks good

PR #209, opened 2024-09-23
example/centos: qcow2 for x86 and aarch64

supakeen:
Initial setup and dedup to build the rest on.


example/centos/common/repositories.yaml
achilleas-k, line 1: How are we handling architectures? Are we going to have separate files (in which case, this should probably be named `repositories-x86_64.yaml`) or have keys that we use to index?

test/data/images-ref/centos/9/aarch64/qcow2/centos_9-aarch64-qcow2-empty.json
achilleas-k, line 1: Eventually, we'll remove these and generate them as needed, right?

achilleas-k (commented):
Looks good in general.  Couple of questions.
But I'd also like to have a nice git history here, I think it will become important to have good context in the future about what's going on here.  It'd be nice to have a separate commit where `test/data/images-ref/` are added with a commit message describing where they came from:

> Generated using `./test/data/images-ref/gen-image-def` for test reference.  In the future, these will be generated on-demand for testing as long as we need them.

It'd also be really nice if changes were broken down a bit by category, like one commit where the `pipeline/build.yaml` file is introduced and and the `otk.include: "pipeline/build.yaml"` is added to the two centos manifests.  Then another commit for the image, etc.

I get this might be a bit cumbersome to split up now, but I also think it's valuable, so feel free do it as far as you think it can be done without it making too much work.

example/centos/pipeline/build.yaml
```diff
@@ -1,21 +1,12 @@
...
-    repositories: ${packages.repositories}
-    gpgkeys: ${packages.keys}
-    packages: ${packages.buildroot.packages}
+  - otk.external.osbuild-make-depsolve-dnf4-rpm-stage:
+      packageset: ${packages.build}
+      gpgkeys:
+        otk.include: "../common/gpgkeys.yaml"
   - type: org.osbuild.selinux
     options:
       file_contexts: etc/selinux/targeted/contexts/files/file_contexts
```
achilleas-k, line 10: We've never had a need to build an image with any other policy (except maybe ISOs?) but it'll be interesting to think about how this would work here.  At some point (not in this PR), we might need to do something to define `selinux_policy="targeted"`, which will affect the path here as well as the package selection for the rpm stage.

And if that's never needed for the selinux policy, the same need will probably arise for something else.  Do we have an idea of how we might do this in a clean way?

example/centos/common/repositories.yaml
mvo5: This also strongly indicates our tests need to catch architectures somehow, we cannot be content with putting everything as "noarch" in the fake depsolver anymore :(

test/test_against_images_refs.py
```diff
@@ -87,6 +88,10 @@ def test_images_ref(tmp_path, monkeypatch, tc):
         ref_manifest = yaml.safe_load(fp)
         normalize_rpm_refs(ref_manifest)
 
+    # hack
+    ref_manifest = json.loads(
```
mvo5, line 92: I feel very uneasy about - it's very broad, all UUIDs will be removed, we have some that we care about like part_type (this will work for them as there is a upper/lower case difference here but relying on this seems risky) - I would really like to explore options and not just yet merge this.

example/centos/common/repositories.yaml
achilleas-k: Using the repo URL to generate the checksum has one obvious downside (the URL becomes significant for the rpm diff, which is not in-line with reality), but it helps catch these kinds of things.

test/test_against_images_refs.py
mvo5: I opened https://github.com/osbuild/otk/pull/212 with a slightly more targeted version (also hacky but in different ways).

test/data/images-ref/centos/9/aarch64/qcow2/centos_9-aarch64-qcow2-empty.json
supakeen: Correct.

example/centos/common/repositories.yaml
mvo5: Indeed, I opened https://github.com/osbuild/images/pull/943 which helps a tiny bit but the real problem is not solved here, i.e. how can we test that the repos that are generated are the same, especially since they are chaning over time

mvo5: One (maybe bad) idea https://github.com/osbuild/images/pull/944 that might help with the second issue.

supakeen:
@achilleas-k All commits have been split out separately.

test/test_against_images_refs.py
supakeen: Resolve, will be discussed separately.

supakeen:
We had a chat on Google Meet and I've resolved the outstanding comments related to SELinux and test(ing|ability).

achilleas-k (approved):
Thanks for the breakdown.  Love it!

LGTM

PR #214, opened 2024-09-23
test: x86 ami import

supakeen:
Generated using `./test/data/images-ref/gen-image-def` for test reference. In the future, these will be generated on-demand for testing as long as we need them.

### DIFF
--- a/test/data/images-ref/centos/9/x86_64/ami/centos_9-x86_64-ami-empty.json
+++ b/test/data/images-ref/centos/9/x86_64/ami/centos_9-x86_64-ami-empty.json
@@ -1,0 +1,754 @@
+{
+  "version": "2",
+  "pipelines": [
+    {
+      "name": "build",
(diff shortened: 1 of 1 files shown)


mvo5 (commented):
We alreeady have the yaml in there in https://github.com/osbuild/otk/tree/main/test/data/images-ref/centos/9/x86_64/ami AFAICT, why do we re-import the json again? What am I missing?

supakeen:
> We alreeady have the yaml in there in https://github.com/osbuild/otk/tree/main/test/data/images-ref/centos/9/x86_64/ami AFAICT, why do we re-import the json again? What am I missing?

I think something weird happened during the splitting out of the previous PRs and it got lost somewhere, this PR only re-adds it.

mvo5 (approved):
Aha, thanks, if it was just lost then of course readding it is fine

PR #215, opened 2024-09-24
external/depsolve: allow passing options for rpm

supakeen:
Allow the passing of options for the RPM stage. Some of our omnifests will need to do this (e.g. `edge-commit` turns off `dracut` and some other bits and bobs).

### DIFF
--- a/src/otk_external_osbuild/command/make_depsolve_dnf4_rpm_stage.py
+++ b/src/otk_external_osbuild/command/make_depsolve_dnf4_rpm_stage.py
@@ -5,7 +5,15 @@ from typing import TextIO
 
 def root(input_stream: TextIO) -> None:
     data = json.load(input_stream)
-    tree = data["tree"]["packageset"]["const"]["internal"]
+    tree = data["tree"]
+    pkgs = tree["packageset"]["const"]["internal"]
+
+    # allow passing on rpm stage options
+    opts = tree.get("options", {}).get("rpm_stage", {})
+
+    # perhaps gpgkeys should move *under* options?
+    # it should: https://github.com/osbuild/otk/issues/218
+    opts["gpgkeys"] = tree["gpgkeys"]
 
     sys.stdout.write(
         json.dumps(
@@ -20,13 +28,11 @@ def root(input_stream: TextIO) -> None:
                                 {
                                     "id": package["checksum"],
                                 }
-                                for package in tree["packages"]
+                                for package in pkgs["packages"]
                             ],
                         },
                     },
-                    "options": {
-                        "gpgkeys": data["tree"]["gpgkeys"],
-                    },
+                    "options": opts,
                 }
             }
         )
--- a/test/test_make_depsolve_dnf4_rpm_stage.py
+++ b/test/test_make_depsolve_dnf4_rpm_stage.py
@@ -58,3 +58,65 @@ def test_make_depsolve_dnf4_rpm_stage(capsys):
             }
         }
     }
+
+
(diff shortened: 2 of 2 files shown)


src/otk_external_osbuild/command/make_depsolve_dnf4_rpm_stage.py
```diff
@@ -5,7 +5,14 @@
...
 def root(input_stream: TextIO) -> None:
     data = json.load(input_stream)
-    tree = data["tree"]["packageset"]["const"]["internal"]
+    tree = data["tree"]
+    pkgs = tree["packageset"]["const"]["internal"]
+
+    # allow passing on stage options
+    opts = tree.get("options", {})
+
+    # perhaps gpgkeys should move *under* options?
```
mvo5, line 14: Yes, I think we should be consistent here, either we expose all rpm stage options under `options` or we hand pick like we did for rpm-keys. Given that our philosophy so far is that we give people the full osbuild stages with all options it seems fine to expose "options" and put gpgkeys under it. It will require a bit of tweaking in the existing manifests but that seems worth the extra consistentcy.

src/otk_external_osbuild/command/make_depsolve_dnf4_rpm_stage.py
```diff
@@ -5,7 +5,14 @@
 
 def root(input_stream: TextIO) -> None:
     data = json.load(input_stream)
-    tree = data["tree"]["packageset"]["const"]["internal"]
+    tree = data["tree"]
+    pkgs = tree["packageset"]["const"]["internal"]
+
+    # allow passing on stage options
+    opts = tree.get("options", {})
```
mvo5, line 12: *maybe* `stage_options` or something (naming is hard) to indicate that this is passed straight to the rpm stage, i.e. that people can just lookup the meta.json to see what is available? But a bit on the fence, options is also nice and short.

mvo5 (commented):
This looks good, thank you! One comment inline and I think we want a simple test (should be trivial given that we have test_make_depsolve_dnf4_rpm_stage.py

src/otk_external_osbuild/command/make_depsolve_dnf4_rpm_stage.py
supakeen: Something to consider in naming/format is that externals can of course generate multiple stages and it won't be entirely clear which stage the options would go to. I picked the `options` name but maybe we should be more explicit and provide explicit key names instead of passing it along (even if it's easy?).

mvo5: Oh, indeed, interesting, I wonder if we should nest then, like 
```
options:
 org.osbuild.rpm: 
   foo: bar
```
or just (because we know what stages are under the external)
```
options:
 rpm_stage:
...
````
but maybe more explicit is actually better here.

supakeen: I prefer the second option. I'll update this PR.

supakeen: Done. Split out the move of `gpgkeys` to #218 to not lose it.

mvo5: Cool, thanks! I think we still want a simple test and then this is good to go from my PoV

mvo5 (approved):
Thanks!

PR #217, opened 2024-09-24
example: x86 and aarch64 for image-installer

supakeen:
Generated image references and deduplicated omnifests.

### DIFF
--- a/example/centos/centos-9-aarch64-image-installer.yaml
+++ b/example/centos/centos-9-aarch64-image-installer.yaml
@@ -375,6 +375,7 @@ otk.target.osbuild:
             size: '5368709120'
         - type: org.osbuild.mkfs.ext4
           options:
+            # we use hardcoded uuids for compatibility with "images"
             uuid: 2fe99653-f7ff-44fd-bea8-fa70107524fb
             label: Anaconda
           devices:
--- a/example/centos/centos-9-aarch64-image-installer.yaml
+++ b/example/centos/centos-9-aarch64-image-installer.yaml
@@ -1,6 +1,12 @@
 otk.version: "1"
 
 otk.define:
+  # XXX these sizes are currently hardcoded in `images` in the future they
+  # XXX might come from an external or elsewhere?
+  # XXX https://github.com/osbuild/images/blob/v0.89.0/pkg/image/anaconda_tar_installer.go#L24
+  size_iso_rootfs: '5368709120'
+  size_iso_efiboot: '20971520'
+  size_iso_efiboot_in_sectors: 40960
   filesystem:
     modifications:
     # empty
@@ -372,7 +378,7 @@ otk.target.osbuild:
         - type: org.osbuild.truncate
           options:
             filename: /LiveOS/rootfs.img
-            size: '5368709120'
+            size: ${size_iso_rootfs}
         - type: org.osbuild.mkfs.ext4
           options:
             # we use hardcoded uuids for compatibility with "images"
@@ -484,7 +490,7 @@ otk.target.osbuild:
         - type: org.osbuild.truncate
           options:
             filename: images/efiboot.img
-            size: '20971520'
+            size: ${size_iso_efiboot}
         - type: org.osbuild.mkfs.fat
           options:
             volid: c81236c2
@@ -493,7 +499,7 @@ otk.target.osbuild:
               type: org.osbuild.loopback
               options:
                 filename: images/efiboot.img
-                size: 40960
+                size: ${size_iso_efiboot_in_sectors}
                 lock: true
         - type: org.osbuild.copy
           inputs:
@@ -511,7 +517,7 @@ otk.target.osbuild:
               type: org.osbuild.loopback
               options:
                 filename: images/efiboot.img
-                size: 40960
+                size: ${size_iso_efiboot_in_sectors}
           mounts:
             - name: '-'
               type: org.osbuild.fat
--- a/example/centos/centos-9-aarch64-image-installer.yaml
+++ b/example/centos/centos-9-aarch64-image-installer.yaml
@@ -1,0 +1,572 @@
+otk.version: "1"
+
+otk.define:
+  filesystem:
+    modifications:
+    # empty
+  packages:
+    build:
+      otk.external.osbuild-gen-depsolve-dnf4:
+        architecture: aarch64
+        module_platform_id: c9s
+        releasever: "9"
+        repositories:
+          otk.include: "common/repositories/aarch64.yaml"
+        packages:
+          include:
+            - coreutils
+            - efibootmgr
+            - glibc
+            - grub2-efi-aa64
+            - grub2-efi-aa64-cdboot
+            - grub2-tools
+            - grub2-tools-extra
+            - grub2-tools-minimal
+            - isomd5sum
+            - lorax-templates-generic
+            - platform-python
+            - policycoreutils
+            - python3
+            - rpm
+            - selinux-policy-targeted
+            - shim-aa64
+            - squashfs-tools
+            - systemd
+            - tar
+            - xorriso
+            - xz
+          exclude: []
+    os:
+      otk.external.osbuild-gen-depsolve-dnf4:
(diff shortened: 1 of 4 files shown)


example/centos/centos-9-aarch64-image-installer.yaml
```diff
@@ -0,0 +1,572 @@
...
+      build: name:build
+      stages:
+        - type: org.osbuild.mkdir
+          options:
+            paths:
+              - path: /LiveOS
+        - type: org.osbuild.truncate
+          options:
+            filename: /LiveOS/rootfs.img
+            size: '5368709120'
```
mvo5, line 375: Hm, this is curious, is the size encoded here as a string? is this as it is supposed to be? Should this be a modification with a default instead maybe? Or just something in an otk.define at least so that we have an easy place to update?

example/centos/centos-9-aarch64-image-installer.yaml
```diff
@@ -0,0 +1,572 @@
...
+        - type: org.osbuild.mkfs.ext4
+          options:
+            uuid: 2fe99653-f7ff-44fd-bea8-fa70107524fb
```
mvo5, line 378: This is also an interesting case, it seems images is hardcoding this, might be worth adding a comemnt

example/centos/centos-9-aarch64-image-installer.yaml
```diff
@@ -0,0 +1,572 @@
...
+              name: CentOS Stream
+              version: 9-stream
+            kernel:
+              dir: /images/pxeboot
+              opts:
+                - inst.stage2=hd:LABEL=CentOS-Stream-9-BaseOS-aarch64
+                - inst.ks=hd:LABEL=CentOS-Stream-9-BaseOS-aarch64:/osbuild.ks
+            isolabel: CentOS-Stream-9-BaseOS-aarch64
+            architectures:
+              - AA64
```
mvo5, line 421: This is also interesting - so `grub2` has it's own set of architecture codes?

example/centos/centos-9-aarch64-image-installer.yaml
```diff
@@ -0,0 +1,572 @@
...
+              type: org.osbuild.tree
+              origin: org.osbuild.pipeline
+              references:
+                - name:rootfs-image
+          options:
+            filename: images/install.img
+            compression:
+              method: xz
+              options:
+                bcj: arm
```
mvo5, line 482: Woah and one more arch identifier.

example/centos/centos-9-aarch64-image-installer.yaml
```diff
@@ -0,0 +1,572 @@
...
+        - type: org.osbuild.truncate
+          options:
+            filename: images/efiboot.img
+            size: '20971520'
```
mvo5, line 486: Similar question about size as string vs number here as above and a bit of a meta question if we need an external or tweak the existing partition external for this block here, it seems it's doing a bunch of size adjustments that seem more suitable to an external(?)

mvo5 (commented):
Thank you, this looks nice. I have a bunch of questions related to the partitioning, I wonder if we need a new helper here or extend the existing one(?)

example/centos/centos-9-aarch64-image-installer.yaml
supakeen: It seems like it is; this is how it was exported directly by `gen-images-ref` at least. Looking at the stage options for `org.osbuild.truncate` it has type `string` for `size`.

example/centos/centos-9-aarch64-image-installer.yaml
supakeen: Yes.

example/centos/centos-9-aarch64-image-installer.yaml
supakeen: Yes :(

example/centos/centos-9-aarch64-image-installer.yaml
supakeen: To say 'needs' to be hardcoded?

mvo5: Well, hardcoding filesystem UUIDs is something unusual, I think it deserves a comment and we also do that in e.g. https://github.com/osbuild/otk/blob/main/example/centos/common/partition-table/x86_64.yaml#L24

So maybe something similar here: `# we use hardcoded uuids for compatibility with "images"` ?

supakeen: Addressed in separate commit.

example/centos/centos-9-aarch64-image-installer.yaml
supakeen: Moved into `${size_rootfs}`.

example/centos/centos-9-aarch64-image-installer.yaml
supakeen: Moved into `${size_efiboot}` with a comment mentioning that it might need an external.

example/centos/centos-9-aarch64-image-installer.yaml
```diff
@@ -0,0 +1,576 @@
+otk.version: "1"
+
+otk.define:
+  # XXX these sizes might need to come from an external?
+  size_rootfs: '5368709120'
```
mvo5, line 5: Maybe `size_iso_rootfs` or `size_iso_rootfs_image` or something to be a) very clear that this is not the disk image rootfs size but the iso and b) to be close to https://github.com/osbuild/images/blob/v0.89.0/pkg/image/anaconda_tar_installer.go#L157  which seems where this is coming from (maybe even ref it here?)

example/centos/centos-9-aarch64-image-installer.yaml
```diff
@@ -0,0 +1,576 @@
...
+  size_efiboot: '20971520'
```
mvo5, line 6: Maybe a comment like `# hardcoded in images` or something and a link to https://github.com/osbuild/images/blob/v0.89.0/pkg/image/anaconda_tar_installer.go#L24 here ?

example/centos/centos-9-aarch64-image-installer.yaml
```diff
@@ -0,0 +1,576 @@
...
+            size: ${size_efiboot}
+        - type: org.osbuild.mkfs.fat
+          options:
+            volid: c81236c2
+          devices:
+            device:
+              type: org.osbuild.loopback
+              options:
+                filename: images/efiboot.img
+                size: 40960
```
mvo5, line 499: We should probably put this value into a define next to `size_efiboot` too - it's `size_efiboot` in sectors so we should add a comment there and then ideally look if we can use the partition table external to generate all of this, it seems to be an extermly simple partition table as defined in https://github.com/osbuild/images/blob/v0.89.0/pkg/image/anaconda_tar_installer.go#L25 and with the defines generated with that we could use `otk-make-partition-stages` and below use `otk-partitions-mounts-devices`.

Given that we have no mechanism right now in images to changes these sizes we could of course just extract a define and add a comment how to calculate the right sizes by hand. But it would be nice to explore if we can do this the external (and maybe tweak/fix as needed). To me at least it feels a bit like starting with tech-debt if we ignore this and when later we copy the image definitions to create e.g. rhel10 and need to change sizes we will need to do calculations like this by hand.

mvo5 (commented):
Sorry, some more comments, I know it's annoying but then this is a new "class" of image so a slightly higher investment compared to the existing ones is not unexpected.

example/centos/centos-9-aarch64-image-installer.yaml
supakeen: Amended the last commit to do this (made the variable name verbose ;)).

example/centos/centos-9-aarch64-image-installer.yaml
supakeen: Amended last commit with new names.

mvo5 (approved):

PR #220, opened 2024-09-24
Makefile: implement proper error message on missing binary

schuellerf:
Added a small help text to get directions as the binaries have to be installed with pip to work, right?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,11 +19,33 @@ type: check-pre-commit
 format:
 	@find src test -name '*.py' | xargs autopep8 --in-place
 
+OSBUILD_GEN_DEPSOLVE_DNF4_EXECUTABLE := $(shell (which "osbuild-gen-depsolve-dnf4" 2>/dev/null))
+OSBUILD_MAKE_DEPSOLVE_DNF4_RPM_STAGE_EXECUTABLE := $(shell (which "osbuild-make-depsolve-dnf4-rpm-stage" 2>/dev/null))
+OSBUILD_MAKE_DEPSOLVE_DNF4_CURL_SOURCE_EXECUTABLE := $(shell (which "osbuild-make-depsolve-dnf4-curl-source" 2>/dev/null))
+
+.PHONY: check_binaries
+check_binaries:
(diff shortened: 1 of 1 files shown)


schuellerf:
btw should the `external` directory stay? (thus should be in `.gitignore` ?)

supakeen:
It can go into `.gitignore`, yes.

Makefile
```diff
@@ -19,11 +19,33 @@ type: check-pre-commit
 format:
 	@find src test -name '*.py' | xargs autopep8 --in-place
 
+OSBUILD_GEN_DEPSOLVE_DNF4_EXECUTABLE := $(shell (which "osbuild-gen-depsolve-dnf4" 2>/dev/null))
+OSBUILD_MAKE_DEPSOLVE_DNF4_RPM_STAGE_EXECUTABLE := $(shell (which "osbuild-make-depsolve-dnf4-rpm-stage" 2>/dev/null))
+OSBUILD_MAKE_DEPSOLVE_DNF4_CURL_SOURCE_EXECUTABLE := $(shell (which "osbuild-make-depsolve-dnf4-curl-source" 2>/dev/null))
+
+.PHONY: check_binaries
+check_binaries:
```
supakeen, line 27: I understand what it's doing and why it's only for the Python ones as the Go ones are already a dependency of the `test` target but I'd like it more to have `check_binaries` be a dependency of `external` (which is called by `test`).

schuellerf: No reason - I just started to understand what's happening here...
actually I would want to have it build it straight forward, if missing. How `make` should work anyhow
I just changed it from "building" it to just an error message as I wasn't sure if I should force people to use the local `venv`

But if you are fine with it I'll directly implement

`venv` missing -> `python3 -m venv venv`
-> always `. venv/bin/activate` in the makefile
`external/*` missing -> `pip install …`

supakeen: It's an OK approach though you can drop the activation in the Makefile and use explicit paths instead.

schuellerf:
obsolete due to #225

PR #221, opened 2024-09-24
README.md: fix documentation of `otk` call

schuellerf:
the "command" (like `compile`) is mandatory and should stay mandatory, right?

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ You can find `otk`'s documentation in the [/doc](./doc) subdirectory. This READM
 If you want to quickly run and try out `otk` without installation the easiest is to run our container image:
 
 ```
-podman run -i ghcr.io/osbuild/otk < omnifest.yaml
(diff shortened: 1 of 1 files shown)


supakeen:
It is, but it is probably better to remove that section as it's broken while we figure out how to do externals and it can't deal with includes (which aren't in the container).

schuellerf:
just "mount" includes with `--volume` ?
also for externals - they are searched in multiple paths anyway - so one can be bind-mounted? (e.g. `./external` -> `/usr/local/libexec/otk/external`)

supakeen:
> just "mount" includes with `--volume` ? also for externals - they are searched in multiple paths anyway - so one can be bind-mounted? (e.g. `./external` -> `/usr/local/libexec/otk/external`)

Sure, this can be done. Do you want to do so in this PR?

schuellerf:
I'll update accordingly, after the outcome of #222

schuellerf:
documentation also handled in #222

PR #211, opened 2024-09-23
example/centos: ami for aarch64

supakeen:
Based on #210.

PR #222, opened 2024-09-25
Containerfile: include externals in the container

schuellerf:
Does it make sense to create a container for now which "just works"?
(related to #221)

This version of the container includes `otk` and all external commands to have it working for now.

build with
```
podman build --tag otk --pull=newer .
```

run e.g. with

```
podman run --rm -ti -v .:/app otk:latest compile /app/example/centos/centos-9-x86_64-tar.yaml
```

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -1,10 +1,9 @@
-FROM registry.access.redhat.com/ubi9/ubi-minimal AS build
+FROM registry.fedoraproject.org/fedora:latest AS build_python
 
-RUN \
-    microdnf install -y \
+RUN dnf install -y \
         python3-pip \
         python3-pyyaml \
-        && microdnf clean all
+        && dnf clean all
 
 WORKDIR /src
 
@@ -12,6 +11,36 @@ COPY pyproject.toml /src/
 COPY src /src/src
 
 RUN pip install --no-dependencies .
-RUN rm -rf src/*
+
+# ----
+FROM registry.access.redhat.com/ubi9/go-toolset:latest AS build_go
+
+ARG CONTAINERS_STORAGE_THIN_TAGS="containers_image_openpgp exclude_graphdriver_btrfs exclude_graphdriver_devicemapper"
+ENV CONTAINERS_STORAGE_THIN_TAGS=${CONTAINERS_STORAGE_THIN_TAGS}
+
+ARG IMAGES_REF="github.com/osbuild/images"
+ENV IMAGES_REF=${IMAGES_REF}
+
+COPY Makefile .
+
+RUN make external
+
+# ----
+# no osbuild-depsolve-dnf here:
+# FROM registry.access.redhat.com/ubi9/ubi-minimal
+FROM registry.fedoraproject.org/fedora:latest
+
+RUN dnf install -y \
+    python3-pyyaml \
+    osbuild-depsolve-dnf \
+    && dnf clean all
+
+COPY --from=build_python /usr/local/bin/otk /usr/local/bin/
+COPY --from=build_python /usr/local/bin/osbuild* /usr/libexec/otk/external/
+COPY --from=build_python /usr/local/lib/ /usr/local/lib/
+
+COPY --from=build_go /opt/app-root/src/external/* /usr/local/libexec/otk/external/
+
+WORKDIR /app
 
 ENTRYPOINT ["otk"]
--- a/Containerfile
+++ b/Containerfile
@@ -15,12 +15,6 @@ RUN pip install --no-dependencies .
 # ----
 FROM registry.access.redhat.com/ubi9/go-toolset:latest AS build_go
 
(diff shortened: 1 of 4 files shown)


schuellerf:
although 
```
podman run --rm -ti -v .:/app otk:latest compile /app/example/centos/centos-9-aarch64-ami.yaml
```
of #211 fails with

```
  File "/usr/local/lib/python3.12/site-packages/otk/external.py", line 32, in call
    raise ExternalFailedError(msg)
otk.error.ExternalFailedError: call: PosixPath('/usr/libexec/otk/external/osbuild-gen-depsolve-dnf4') 'otk.external.osbuild-gen-depsolve-dnf4' failed: '', 'Traceback (most recent call last):\n  File "/usr/libexec/otk/external/osbuild-gen-depsolve-dnf4", line 8, in <module>\n    sys.exit(main())\n             ^^^^^^\n  File "/usr/local/lib/python3.12/site-packages/otk_external_osbuild/command/gen_depsolve_dnf4.py", line 86, in main\n    root(sys.stdin)\n  File "/usr/local/lib/python3.12/site-packages/otk_external_osbuild/command/gen_depsolve_dnf4.py", line 77, in root\n    raise RuntimeError(f"{process.stdout=}{process.stderr=}")\nRuntimeError: process.stdout=\'{"kind": "MarkingErrors", "reason": "Error occurred when marking packages for installation: Problems in request:\\\\nmissing packages: platform-python"}\\n\'process.stderr=\'No match for argument: platform-python\\nerror install_specs\\nMarkingErrors: Error occurred when marking packages for installation: Problems in request:\\nmissing packages: platform-python\\n\'\n'
```

Containerfile
```diff
@@ -1,17 +1,52 @@
...
+
+ARG CONTAINERS_STORAGE_THIN_TAGS="containers_image_openpgp exclude_graphdriver_btrfs exclude_graphdriver_devicemapper"
+ENV CONTAINERS_STORAGE_THIN_TAGS=${CONTAINERS_STORAGE_THIN_TAGS}
+
+ARG IMAGES_REF="github.com/osbuild/images"
+ENV IMAGES_REF=${IMAGES_REF}
+
+# Note that "external" will most likely in the future build from internal
+# sources instead of pulling off the network
+RUN sh -c 'for otk_cmd in gen-partition-table \
```
mvo5, line 26: Silly question, could we just run "make externals" or something here to avoid the duplication of this external list (and the exclude tags)?

schuellerf: Sure, thanks for the hint 🥳, I would have optimized this once it's fully working (see my last comment) and I got the confirmation that we should even package this into the container (in contrast to bind-mounting it)

Where I think a container which is fully working (including all externals we know) is beneficial to get started faster

mvo5: Oh, sorry, then my comment was a bit premature I guess (sry!)

supakeen:
Yes; it makes sense to have a container which 'just works' :)

schuellerf:
btw I had to switch away from `ubi9/ubi-minimal` to fedora as base image due to `osbuild-depsolve-dnf` not being available - is this correct, or can this be fixed somehow

schuellerf:
> although CONTAINER_TEST_FILE=example/centos/centos-9-aarch64-ami.yaml make container-test does not work

That might be related to cross-arch building... (the x86_64 versions are ok)
So this might be fine for this PR

supakeen (approved):

Containerfile
```diff
@@ -1,17 +1,46 @@
...
 COPY pyproject.toml /src/
 COPY src /src/src
 
 RUN pip install --no-dependencies .
-RUN rm -rf src/*
+
+# ----
+FROM registry.access.redhat.com/ubi9/go-toolset:latest AS build_go
+
+ARG CONTAINERS_STORAGE_THIN_TAGS="containers_image_openpgp exclude_graphdriver_btrfs exclude_graphdriver_devicemapper"
```
mvo5, line 18: Can we somehow avoid duplicating this? It's already in the Makefile iirc?

schuellerf: I wanted to do that but I'm not sure how to do that. For me it's better to have a sane fallback if someone builds this `Containerfile` directly.
having some grep here and there or accepting that `Containerfile` can't be built without that knowledge seems both bad to me

schuellerf: grepping from the Containerfile into the Makefile could work - if that's nicer? 🤷‍♂️

mvo5: Sorry, maybe I'm just confused by we call `make` in the Containerfile, why is it not picking up the options that we set in the makefile, i.e. why do we need them here again? And sorry for being to nitpicky about this but this list irks me a great deal already, we have it in bootc-image-builder too and now we have it here in two places, I would really like to minimize the places we need to update :)

schuellerf: 🫣 to no surprise, you are right!
This was due to refactoring (make was not called before) and totally not needed… Thanks!

supakeen (approved):

Issue #201, opened 2024-09-16
Figure out a better way to test externals


supakeen:
With the way we do external paths and the src-layout for the Python package combined with the Go externals it's a bit cumbersome to test things.

I created a stopgap to `make external` in #196, let's replace it with something better.

PR #223, opened 2024-09-25
Makefile: make python externals available without venv

mvo5:
To run the pyton externals currently a venv is needed. However things work fine without except for this.

This commit tweaks the way the externals are made available so that no venv is needed by just creating small wrapper binaries.

Followup for https://github.com/osbuild/otk/pull/204 Closes: https://github.com/osbuild/otk/issues/201 (at least for me)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,9 +21,10 @@ format:
 
 .PHONY: test
 test: external
-	cp $(shell (which "osbuild-gen-depsolve-dnf4")) ./external/
-	cp $(shell (which "osbuild-make-depsolve-dnf4-rpm-stage")) ./external/
-	cp $(shell (which "osbuild-make-depsolve-dnf4-curl-source")) ./external/
+	echo '#!/usr/bin/python3 -motk_external_osbuild.command.gen_depsolve_dnf4' > ./external/osbuild-gen-depsolve-dnf4
+	echo '#!/usr/bin/python3 -motk_external_osbuild.command.make_depsolve_dnf4_rpm_stage' > ./external/osbuild-make-depsolve-dnf4-rpm-stage
+	echo '#!/usr/bin/python3 -motk_external_osbuild.command.make_depsolve_dnf4_curl_source'> ./external/osbuild-make-depsolve-dnf4-curl-source
+	chmod +x external/*
 	@pytest
 
 .PHONY: push-check


Makefile
```diff
@@ -21,9 +21,10 @@ format:
 
 .PHONY: test
 test: external
-	cp $(shell (which "osbuild-gen-depsolve-dnf4")) ./external/
-	cp $(shell (which "osbuild-make-depsolve-dnf4-rpm-stage")) ./external/
-	cp $(shell (which "osbuild-make-depsolve-dnf4-curl-source")) ./external/
+	echo '#!/usr/bin/python3 -motk_external_osbuild.command.gen_depsolve_dnf4' > ./external/osbuild-gen-depsolve-dnf4
+	echo '#!/usr/bin/python3 -motk_external_osbuild.command.make_depsolve_dnf4_rpm_stage' > ./external/osbuild-make-depsolve-dnf4-rpm-stage
+	echo '#!/usr/bin/python3 -motk_external_osbuild.command.make_depsolve_dnf4_curl_source'> ./external/osbuild-make-depsolve-dnf4-curl-source
```
supakeen, line 26: This still requires `otk_external_osbuild` to be installed somewhere on an importable Python path which won't be by default in the repository root right?

mvo5: It requires PYTHONPATH=./src or similar when outside the venv and it inside it works except it does not :(
```
mvogt@top:~/devel/osbuild/otk$ python3 -m venv venv
mvogt@top:~/devel/osbuild/otk$  . venv/bin/activate
(venv) mvogt@top:~/devel/osbuild/otk$ pip install -e ".[dev]"
...
$ echo {}|python3 -motk_external_osbuild.command.gen_depsolve_dnf4
...
KeyError: 'tree'
$ make test
...
E           otk.error.ExternalFailedError: call: PosixPath('external/osbuild-gen-depsolve-dnf4') 'otk.external.osbuild-gen-depsolve-dnf4' failed: '', "/usr/bin/python3: Error while finding module specification for 'otk_external_osbuild.command.gen_depsolve_dnf4' (ModuleNotFoundError: No module named 'otk_external_osbuild')\n"
```
apparently because VIRTUAL_ENV is stripped from the env.

I will close this again as clearly something that does not work inside a venv is silly but I'm also a bit annoyed at python for requiring this byzantian setup for binaries (which I suspect is needed for windows and we pay for this complexity). I would really love to just have a `./src/otk/externals/osbuild-gen-depsolve-dnf4` with:
```
#!/usr/bin/env python3
from otk_external_osbuild.command.gen_depsolve_dnf4 import main
if __name__ == "__main__":
    main()
```
afaict this would just work in venv or with a set pythonpath. oh well.

PR #224, opened 2024-09-25
test: drop the image-ref json files as we have the yaml

mvo5:
While reviewing I noticed that we have the image_ref json files for all our tests. This seems unnecessary given that we have the yaml and we use the yaml in the tests to compare generated with reference.

So this commit just drops the json references (and future PRs should not include include them if we merge this one).


supakeen (approved):
Might need adjustment to other outstanding PRs.

PR #225, opened 2024-09-25
Implement makefile building from scratch

schuellerf:
This fixes the Makefile and introduces new targets.
e.g. `make test` now also sets up the venv and externals
Now the following rules should be working:

* help: new
* clean: new - be aware this also removes venv as it will be rebuilt anyway
* test: did not setup the environment before
* rpm: was failing in the current version

also updates `.gitignore` to match (so [this](https://github.com/osbuild/otk/pull/220#issuecomment-2371727229) is obsolete)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,5 @@ htmlcov
 dictionary.dic
 demo
 /container_built.info
+/external
+/rpmbuild
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,33 @@
+# needed for venv
+.ONESHELL:
 
-check-pre-commit:
+.PHONY: help
+help:
+	@echo 'Usage:'
+	@echo '  make <target>'
+	@echo ''
+	@echo 'Targets:'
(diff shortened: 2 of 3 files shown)


Containerfile
```diff
@@ -1,17 +1,52 @@
...
+
+ARG CONTAINERS_STORAGE_THIN_TAGS="containers_image_openpgp exclude_graphdriver_btrfs exclude_graphdriver_devicemapper"
+ENV CONTAINERS_STORAGE_THIN_TAGS=${CONTAINERS_STORAGE_THIN_TAGS}
+
+ARG IMAGES_REF="github.com/osbuild/images"
+ENV IMAGES_REF=${IMAGES_REF}
+
+# Note that "external" will most likely in the future build from internal
+# sources instead of pulling off the network
+RUN sh -c 'for otk_cmd in gen-partition-table \
```
mvo5, line 26: I will risk another premature comment here but I think we want to use the Makefile and not duplicate this list :)

schuellerf: sorry - my mistake - this commit should not be here

supakeen (approved):

supakeen (approved):

PR #226, opened 2024-09-25
external: add new `get-dnf4-package-info` external

mvo5:
external: add new `get-dnf4-package-info` external

This commit adds a new `get-dnf4-package-info` external that can
be used to extract details about a dnf4 resolved package.

The main use-case is the kernel package where we need some details
like the version number to put into the grub stage. With this
commit we can do something like:
```yaml
  kernel:
    package:
      otk.external.osbuild-get-dnf4-package-info:
        packageset: ${packages.os}
        package_name: ${modifications.kernel_package_name}
```

----

example: use new `osbuild-get-dnf4-package-info` external

This commit uses the new `osbuild-get-dnf4-package-info` for the
real otk examples.

---

external: update mock data in gen_depsolve_dnf4 to follow
 images

This commit updates the generated mock meta-data from the
`gen_depsolve_dnf4` to follow the `gen_manifests` updates from the
images library, most notably:
https://github.com/osbuild/images/pull/937
https://github.com/osbuild/images/pull/943

Alternative idea to https://github.com/osbuild/otk/pull/205 which avoids the issues that we need nested ${packages.${kerel.pkgname}} declarations.


src/otk_external_osbuild/command/get_dnf4_package_info.py
```diff
@@ -0,0 +1,39 @@
...
+def find_pkg_by_name(packages: List[dict], pkg_name: str) -> Optional[dict]:
+    for pkg in packages:
+        if pkg["name"] == pkg_name:
+            return pkg
+    return None
+
+
+def root(input_stream: TextIO) -> None:
+    data = json.load(input_stream)
+    pkg_name = data["tree"]["package_name"]
```
supakeen, line 15: This is going to be EXTREME NITPICKING but `packagename` to be consistent with `packageset`?

supakeen (dismissed):
I think it's a good idea to have an external to do this to dodge other problems indicated in the other PR.

achilleas-k (approved):

src/otk_external_osbuild/command/get_dnf4_package_info.py
mvo5: It does look nicer/more consistent this way, thanks you! I updated the PR to use "packagename" now.

supakeen (approved):

PR #219, opened 2024-09-24
example/centos: x86,aarch64 for vhd, x86 for vmdk

supakeen:
Import more reference manifests and implement deduplicated omnifests. These are lacking the use of the partition external, this will be done in the next commit.

PR #230, opened 2024-09-26
example/centos: qcow2 for ppc64le

schuellerf:
Implements the example of ppc64le support for qcow2

this is based on https://github.com/osbuild/otk/pull/219
so my changes are just the last two commits and this
needs https://github.com/osbuild/images/pull/952 merged first

also I'm not sure if [i should be doing this](https://github.com/osbuild/otk/commit/cdc084e3493ec33e627681a78c90b8dd79b34b49) to patch the UUIDs in the test target?


supakeen:
As I understand it the manual adjustment to the UUIDs should not be necessary anymore. Make sure you're rebased (again) on top of #219. Make sure your images check out is up to date, and so are your externals and then generate the reference manifests and it should (tm) just work.

Do I miss something @mvo5?

schuellerf:
@supakeen  my "images" is acutally up to date and [patched](https://github.com/osbuild/images/pull/952) ;)

schuellerf:
(that's why the tests fail - they need images merged first)

mvo5:
> As I understand it the manual adjustment to the UUIDs should not be necessary anymore. Make sure you're rebased (again) on top of #219. Make sure your images check out is up to date, and so are your externals and then generate the reference manifests and it should (tm) just work.
> 
> Do I miss something @mvo5?

It seems the issue is that ppc64le does not use "label" when creating the filesystems which the gen-images-ref script uses to find the uuids, I can look and see if alternatively it can use the fstab stage (iirc it was doing that originally but then got changed). Manual sed would work in the meantime to unblock you

PR #232, opened 2024-09-27
example/centos: import WSL

supakeen:
Generate test manifests for the WSL image types and implement their initial omnifests.

### DIFF
--- a/example/centos/centos-9-aarch64-wsl.yaml
+++ b/example/centos/centos-9-aarch64-wsl.yaml
@@ -1,0 +1,134 @@
+otk.version: "1"
+
+otk.define:
+  packages:
+    build:
+      otk.external.osbuild-gen-depsolve-dnf4:
+        architecture: aarch64
+        module_platform_id: c9s
+        releasever: "9"
+        repositories:
+          otk.include: "common/repositories/aarch64.yaml"
+        packages:
+          include:
+            - coreutils
+            - glibc
+            - platform-python
+            - python3
+            - python3-iniparse
+            - rpm
+            - selinux-policy-targeted
+            - systemd
+            - tar
+            - xz
+          exclude: []
+    os:
+      otk.external.osbuild-gen-depsolve-dnf4:
+        architecture: aarch64
+        module_platform_id: c9s
+        releasever: "9"
+        repositories:
+          otk.include: "common/repositories/aarch64.yaml"
+        packages:
+          include:
+            - alternatives
+            - audit-libs
+            - basesystem
+            - bash
+            - ca-certificates
+            - coreutils-single
+            - crypto-policies-scripts
+            - curl-minimal
+            - dejavu-sans-fonts
+            - dnf
+            - filesystem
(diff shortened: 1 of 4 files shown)


example/centos/centos-9-aarch64-wsl.yaml
```diff
@@ -0,0 +1,131 @@
...
+          options: {}
+        - type: org.osbuild.locale
+          options:
+            language: en_US.UTF-8
+        - type: org.osbuild.timezone
+          options:
+            zone: America/New_York
+        - type: org.osbuild.sysconfig
+          options:
+            kernel:
```
mvo5, line 103: This is curious, we don't seem to install a kernel but we list it here in sysconfig. Obviously not an issue with this PR as it just replicates images but might still be worth adding a comment (here and in the other file) that we should look into this. I assume this will be the basis of new e.g. centos-10 images where we can look into dropping this kernel option.

achilleas-k: I think this might be an issue in images.  We add the kernel sysconfig to all image types unconditionally across a distro as part of the default image config.  We should probably have some conditions for that.

mvo5: Thanks! If it's an issue with the "images" lib I think it would be good to add a comment here (well, a comment here would be good also if it's fine, I find it confusing otherwise to have a kernel reference here without a kernel :)

supakeen: What do you want the comment to say?

supakeen: I've added the comment verbatim (with typos fixed ;)), but only to the added wsl omnifest here.

mvo5: We want the comment for both files (both arches), no? Or will one file go away RSN with the dedup work?

supakeen: I've added a duplicated comment to the `x86_64` omnifest; I'm not sure how long it will take for this to go away :)

mvo5 (approved):
Thank you

example/centos/centos-9-x86_64-wsl.yaml
```diff
@@ -0,0 +1,245 @@
...
+          options:
+            filename: disk.tar.gz
+          inputs:
+            tree:
+              type: org.osbuild.tree
+              origin: org.osbuild.pipeline
+              references:
+                - name:os
+  sources:
+    org.osbuild.curl:
```
mvo5, line 132: Meh, sorry, I overlooked this :( this needs to be:
```
  sources:
    otk.external.osbuild-make-depsolve-dnf4-curl-source:
      packagesets:
        - ${packages.build}
        - ${packages.os}
```
of course instead of the verbatim copy here

supakeen: #256

PR #238, opened 2024-09-27
Add externals for managing embedded files

achilleas-k:
This PR adds two externals:

**external: add new `gen-inline-files` external**

Will be used to generate inline files for manifests.

To add a file in a manifest as in-line content we need an ID, which is
the sha256 hash of the contents and the file contents in base64.

**external: add new `make-inline-source` external**

Takes the output from gen-inline-files and generates the
org.osbuild.inline source structure for the manifest.

### DIFF
--- a/src/otk_external_osbuild/command/gen_inline_files.py
+++ b/src/otk_external_osbuild/command/gen_inline_files.py
@@ -1,27 +1,43 @@
 import base64
 import hashlib
 import json
+import pathlib
 import sys
 from typing import TextIO
+
+
+def process_contents(contents):
+    digest = hashlib.sha256(contents).hexdigest()
+    source_id = f"sha256:{digest}"
+    b64contents = base64.b64encode(contents).decode("utf8")
+    return {
+        "id": source_id,
+        "data": b64contents,
+    }
 
 
 def root(input_stream: TextIO) -> None:
     data = json.loads(input_stream.read())
     tree = data["tree"]
 
-    files = tree["files"]
+    inline_files = tree.get("inline", {})
     inlines: dict = {}
-    for name, item in files.items():
+    for name, item in inline_files.items():
         contents = item["contents"].encode("utf8")
+        inlines[name] = process_contents(contents)
 
-        digest = hashlib.sha256(contents).hexdigest()
-        source_id = f"sha256:{digest}"
+    inline_paths = tree.get("paths", {})
+    for name, item in inline_paths.items():
+        path = item["path"]
+        read_type = item["type"]
+        if read_type == "text":
+            contents = pathlib.Path(path).read_text(encoding="utf-8").encode("utf-8")
+        elif read_type == "binary":
+            contents = pathlib.Path(path).read_bytes()
+        else:
+            raise ValueError(f"invalid type for inline path {path}: {read_type}")
 
-        b64contents = base64.b64encode(contents).decode("utf8")
-        inlines[name] = {
-            "id": source_id,
-            "data": b64contents,
-        }
+        inlines[name] = process_contents(contents)
 
     output = {
         "tree": {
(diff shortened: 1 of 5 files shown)


supakeen (commented):
I think it needs a bit more, namely `osbuild-make-inline-copy-stage` to also make the copy stage where a file gets used.

When we do that we probably also want a way to refer to these files by a key so I'd say that the data object for `gen-inline-files` should be a map[map] instead of a list[map] where the keys are those identifiers.

```
otk.external.gen-inline-files:
  files:
    simon:
      contents: ...
    not-simon:
      contents: ...
```

As a thought; when we do include files from path would we add another key so that it's either `contents` or `path`? :)

achilleas-k:
> I think it needs a bit more, namely `osbuild-make-inline-copy-stage` to also make the copy stage where a file gets used.
> 
> When we do that we probably also want a way to refer to these files by a key so I'd say that the data object for `gen-inline-files` should be a map[map] instead of a list[map] where the keys are those identifiers.
> 
> ```
> otk.external.gen-inline-files:
>   files:
>     simon:
>       contents: ...
>     not-simon:
>       contents: ...
> ```
> 
> As a thought; when we do include files from path would we add another key so that it's either `contents` or `path`? :)

Are we doing paths?  I guess we should.
But you're right, the main thing would be to have identifiers for files.

As for an inline copy stage, I'm not sure it's necessary.  I mean, a copy stage, when you have identifiers/names for the inline data would just be:
```yaml
otk.define:
  inlines:
    otk.external.gen-inline-files:
      files:
        stuff:
          contents: "abc"

...

type: org.osbuild.copy
inputs:
  file-stuff:
    origin: org.osbuild.source
    references:
    - id: ${inlines.stuff.id}
    type: org.osbuild.files
options:
  paths:
  - from: input://file-stuff/${inlines.stuff.id}
    to: tree:///destination/path/stuff
```

Do we need an external for that?

achilleas-k:
Changed the input/output structure of gen-inline-files from a list to a dictionary.

supakeen:
> Are we doing paths? I guess we should. But you're right, the main thing would be to have identifiers for files.

We will be doing paths, I'm sure of it (it's the most requested feature in blueprints too).

> Do we need an external for that?

Not really for now though it'd be consistent with the other things. We can also add it later when it would generate multiple stages as this external grows to support filemodes and owners (as it likely will).

src/otk_external_osbuild/command/gen_inline_files.py
```diff
@@ -0,0 +1,40 @@
...
+
+        b64contents = base64.b64encode(contents).decode("utf8")
+        inlines[name] = {
+            "id": source_id,
+            "data": b64contents,
+        }
+
+    output = {
+        "tree": {
+            "files": inlines,
```
supakeen, line 28: I guess if we want to be consistent with the other stages this should be `const`ed as it becomes our public API for this external.

supakeen (commented):
Last nitpick for consistency, otherwise <3

src/otk_external_osbuild/command/gen_inline_files.py
achilleas-k: Done.

achilleas-k: Done.

achilleas-k:
> > Are we doing paths? I guess we should. But you're right, the main thing would be to have identifiers for files.
> 
> We will be doing paths, I'm sure of it (it's the most requested feature in blueprints too).

Done.

I'm not entirely dedicated to the structure I came up with to separate the two forms, so feel free to nitpick and bikeshed etc.

supakeen (approved):
I'm happy about it; we can bikeshed the plurality name separately when we go over all the external naming :)

PR #242, opened 2024-09-30
[RFC] example: allow user modifications via `otk.define`

mvo5:
This commit adds support for user modifications. The idea behind modifications is that those are variables that users want to change to customize the imagess.

This appears uses the existing otk.defines mechanism and adds a convention around the naming for modifications.

We will use the customizations from blueprints and map them to modifcations. A simple example would be:
```yaml
otk.define:
 user:
  modifications:
   language: nl_NL.UTF-8
```

In our otk files we will have defaults for all supported modifications and then join the defaults with the user define modifcations.
```yaml
otk.define:
  default:
    modifications:
      language: en_US.UTF-8
  modifications:
    otk.op.join:
      values:
        - ${default.modifications}
        - ${user.modifications}
```

This allows us to follow the idea from the images PR#797 (https://github.com/osbuild/images/pull/797):
```yaml
otk.define:
  user:
    modifications:
      hostname: "hal-9000"
      kernel_append: "debug"
      keyboard: "uk"

otk.include: `/otkroot/fedora-40/x86_64/qcow2.yaml`
```
and because the otk.op.join will merge two tree this should work.

We could also (later) provide a convinience helper like:
```
$ otk compile --preload/--include "my-customizatins.yaml" centos-9-x86_64-ami.yaml
```
which would just preload/include (terms stawman) the customizations and then proceed processing the centos-9-x86_64-ami.yaml file.

To see what modifications are supported by a given image we would just regrep what "modifications" vars are being used (via otk).

---

This PR now also includes a test, the idea is that we generate a reference manifest via images with all the "blueprint" customizations that we support via otk "modifications" and compare the result in an automatic test.

See also https://github.com/osbuild/otk/pull/55#issuecomment-2109938246 (and the whole initial pr#55)


supakeen:
> To see what modifications are supported by a given image we would just regrep what "modifications" vars are being used.

I feel like `otk` should support this usecase by itself otherwise everyone will have to figure out the same regular expression(s).

Otherwise this idea is great and we can always expand on it later, I'd really like if we can have the `--prefix` (that's what it is right?) in this PR as well.

mvo5:
> > To see what modifications are supported by a given image we would just regrep what "modifications" vars are being used.
> 
> I feel like `otk` should support this usecase by itself otherwise everyone will have to figure out the same regular expression(s).

Yeah, sorry, I think I did not explain this well - this knowledge would be encapsulated inside otk and we would provide a CLI interface like `otk info --modifications <path>` or something to extract them. This would also allow us to have more freedom about switching the internal implementation/conventions.
 
> Otherwise this idea is great and we can always expand on it later, I'd really like if we can have the `--prefix` (that's what it is right?) in this PR as well.

Cool, I can certainly add this (once the testing is sorted a bit more).

mvo5:
I updated and resolved the conflicts, I did not add the "--prefix/--prelude/--include" feature and I still would like to get an ACK from @achilleas-k and/or @ondrejbudai before merging/un-drafting.

ondrejbudai (approved):
My only concern is that this having the same glue code in all top-level omnifests will be quite a lot of duplication without much value. I wonder if there's a sane way to deduplicate them (MVP: for disk images, they literally share the same set of customizations).

achilleas-k:
> My only concern is that this having the same glue code in all top-level omnifests will be quite a lot of duplication without much value. I wonder if there's a sane way to deduplicate them (MVP: for disk images, they literally share the same set of customizations).

I assume this can be covered with an `include`?

achilleas-k:
> I updated and resolved the conflicts, I did not add the "--prefix/--prelude/--include" feature and I still would like to get an ACK from @achilleas-k and/or @ondrejbudai before merging/un-drafting.

Just to make sure I understand this correctly:

the idea of a `--prefix` (or whatever we end up calling it) is to point to a file with some specially-keyed `define`s that are preloaded and take precedent over anything that comes from the entrypoint.

supakeen:
> > I updated and resolved the conflicts, I did not add the "--prefix/--prelude/--include" feature and I still would like to get an ACK from @achilleas-k and/or @ondrejbudai before merging/un-drafting.
> 
> Just to make sure I understand this correctly:
> 
> the idea of a `--prefix` (or whatever we end up calling it) is to point to a file with some specially-keyed `define`s that are preloaded and take precedent over anything that comes from the entrypoint.

How I understand it it would be a normal define that goes to the head of the file (hence prefixed). Any merging and priority of those needs to be done manually in the omnifest; for example through the joins in @mvo5's example in this PR.

mvo5:
It seems there is consensus now that this is a reasonable way to do it, so I marked this "ready"

supakeen:
There is, and I've started on an implementation.

PR #243, opened 2024-09-30
src: include current file in every raise()

mvo5:
This commit adds the current file that was processed in all otk exceptions. This way it's easier for the user to spot where something went wrong.

This will now produce errors like:
```
...
otk.error.TransformDirectiveTypeError: [/home/mvogt/devel/osbuild/otk/example/centos/common/partition-table/x86_64/default.yaml] cannot join ['x']
```

[draft as I think this should be two commits and also it lacks dedicated tests, also there should probably a "state.make_error_prefix()" helper or something so that this is all in one place and externals should also be covered but I still wanted to open it so that it's not forgoten]

### DIFF
--- a/src/otk/transform.py
+++ b/src/otk/transform.py
@@ -43,7 +43,7 @@ def resolve(ctx: Context, state: State, data: Any) -> Any:
         return data
 
     log.fatal("could not look up %r in resolvers", type(data))
-    raise TypeError(type(data))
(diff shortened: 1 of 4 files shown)


mvo5:
This is no longer needed, it got solved in a cleaner way via https://github.com/osbuild/otk/pull/263 and https://github.com/osbuild/otk/pull/265

PR #244, opened 2024-09-30
otk: do not allow duplicated yaml keys

mvo5:
This commit will make the yaml loader error when duplicated yaml keys are used. This can easily happen with e.g.:
```yaml
otk.include: foo.yaml
...
[lots of stuff]
...
otk.include: bar.yaml
```
with the default yaml loader this would make the value of `otk.include=bar.yaml` which leads to confusing and hard to understand errors. Best is to simply error with a reasonable error message.

The correct way in otk to write this would be:
```
otk.include.foo: foo.yaml
...
otk.include.bar: bar.yaml
```

This will now yield errors like:
```
...
file .../otk/test/data/error/09-duplicated-key.yaml has duplicated yaml keys: duplicate 'otk.include' key found, try using otk.include.<uniq-tag>, e.g. otk.include.foo
```

Note that this PR also removes one of the `yaml` linters (`check-yaml`), see the individual commit message for details.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -8,7 +8,6 @@ repos:
         args:
           - --markdown-linebreak-ext=md
       - id: end-of-file-fixer
-      - id: check-yaml
       - id: check-added-large-files
   - repo: https://github.com/adrienverge/yamllint
     rev: v1.35.1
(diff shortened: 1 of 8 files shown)


supakeen (approved):
Agreed on the behavioral change. I'd like it if you linked to the YAML specification that disallows non-unique mapping keys (after resolution) as well: https://yaml.org/spec/1.2.2/#node-comparison

supakeen (approved):

PR #245, opened 2024-09-30
example/centos: edge-raw for x86_64

achilleas-k:
This addition has a few issues:
- We need LUKS+LVM support in the partition table generator.
- We need a new external for generating the ostree source list: https://github.com/osbuild/images/pull/960
- We might have to support resolving variables in object keys.

See individual commit messages and comments for more info.

Draft until new external lands.


schuellerf:
project will be archived…

PR #246, opened 2024-10-01
example/centos: drop unused files

supakeen:
Drops all unused fragments and the common centos 9 as they are not used currently. This will allow me to build them back up again.

Split out from #240.


mvo5 (approved):
Thank you!

PR #247, opened 2024-10-01
example/centos: locale fragment

supakeen:
Introduce a locale fragment that fills in an osbuild locale stage with a global define.

Note that the image-installer omnifests are excluded as they currently have dual locales so can't use a single global define.

Split out (and updated) from #240.

### DIFF
--- a/example/centos/centos-9-aarch64-ami.yaml
+++ b/example/centos/centos-9-aarch64-ami.yaml
@@ -1,6 +1,8 @@
 otk.version: "1"
 
 otk.define:
+  modifications:
+    locale: en_US.UTF-8
   filesystem:
     modifications:
       filename: "image.raw"
@@ -120,9 +122,7 @@ otk.target.osbuild:
         - type: org.osbuild.fix-bls
           options:
             prefix: ''
-        - type: org.osbuild.locale
-          options:
-            language: en_US.UTF-8
+        - otk.include: "fragment/locale.yaml"
         - type: org.osbuild.keymap
           options:
             keymap: us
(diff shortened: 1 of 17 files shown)


supakeen:
An interesting question here, maybe related to #242 is that I've left out the image installer types initially as they contain two *different* locales (which probably is a 'bug'?) however, how would we do the modification convention when that case does arise? (cc: @mvo5).

mvo5 (approved):
Thank you, this looks very nice and clean. 

About the image installer types - I had a quick look at images and it seems the "C.UTF-8" is the one that is influenced by a customization, we should probably just follow suite here, i.e. in both:
```yaml
diff --git a/test/data/images-ref/centos/9/x86_64/image-installer/centos_9-x86_64-image_installer-empty.yaml b/test/data/images-ref/centos/9/x86_64/image-installer/centos_9-x86_64-image_installer-empty.yaml
index e855e17..0c8595c 100644
--- a/test/data/images-ref/centos/9/x86_64/image-installer/centos_9-x86_64-image_installer-empty.yaml
+++ b/test/data/images-ref/centos/9/x86_64/image-installer/centos_9-x86_64-image_installer-empty.yaml
@@ -673,7 +673,7 @@ pipelines:
         options: {}
       - type: org.osbuild.locale
         options:
-          language: C.UTF-8
+          language: ${modifications.locale}
       - type: org.osbuild.timezone
         options:
           zone: America/New_York
``` 
and then a comment in the other place like `# anaconda tree is always using en_US.UTF-8 currently, see https://github.com/osbuild/images/blob/v0.90.0/pkg/manifest/anaconda_installer.go#L239` (or something like this)

PR #248, opened 2024-10-01
example/centos: fix-bls fragment

supakeen:
Introduce a new fix-bls fragment that comes in two variants, one with an empty prefix and one with the default options selected. Perhaps this can be further deduplicated later on (doing this based on a modification?).

Split out from #240.

### DIFF
--- a/example/centos/fragment/fix-bls/default.yaml
+++ b/example/centos/fragment/fix-bls/default.yaml
@@ -1,0 +1,6 @@
+# If `/boot` is on a separate partition the `empty` BLS fragment should be used, otherwise the default version
+# is used.
+#
+# See: https://github.com/osbuild/images/blob/v0.90.0/pkg/manifest/os.go#L439
+type: org.osbuild.fix-bls
+options: {}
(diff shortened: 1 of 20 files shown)


example/centos/fragment/fix-bls/default.yaml
```diff
@@ -0,0 +1,2 @@
+type: org.osbuild.fix-bls
+options: {}
```
mvo5, line 2: I think it would be good if both of these fragments have a comment that point to https://github.com/osbuild/images/blob/v0.90.0/pkg/manifest/os.go#L439 and explains a little bit what the rules for adding a BLS stage are (and also it's one thing to think about, i.e. here it is nice that images is code as it can make this decision automatically (cc @ondrejbudai @achilleas-k ) - it would be nice to think if there is a way to provide some user safeguards (an external seems overkill but might work)

supakeen: It seems to me that perhaps the creation of the fix-bls stage could be handled by the partition external? That one knows if `/boot` is on a (separate) partition.

mvo5 (approved):

example/centos/fragment/fix-bls/default.yaml
mvo5: I like the idea of re-using the partition external

supakeen: https://github.com/osbuild/otk/issues/252

PR #249, opened 2024-10-01
example/centos: timezone fragment

supakeen:
Introduce a new timezone fragment. Split off from #240.


example/centos/centos-9-aarch64-vhd.yaml
```diff
@@ -3,6 +3,7 @@ otk.version: "1"
 otk.define:
   modifications:
     locale: en_US.UTF-8
+    timezone: Etc/UTC
```
mvo5, line 6: quite interesting how much variance we have in the timezones :)

mvo5 (approved):
Thank you

PR #254, opened 2024-10-01
test: use `fstab` stage in `gen-image-def` to find UUIDs

mvo5:
This commit changes the `gen-image-def` script to use the `org.osbuild.fstab` stage to find the UUIDs of the root and boot partitions. We used to use the `mkfs.*` stage and the `label` parameter but it turns out that `s390x` and `ppc64le` do not use the label paramter so the UUID is not found.

Thanks to Florian Schüller <florian.schueller@redhat.com> for finding and testing.


schuellerf (approved):
thanks alot!
worked for my s390x usecase 🥳

PR #255, opened 2024-10-01
example/centos: introduce `${architecture}`

supakeen:
A variable containing the targeted architecture makes omnifests less error prone to write. Split out from #240.

### DIFF
--- a/example/centos/centos-9-aarch64-edge-commit.yaml
+++ b/example/centos/centos-9-aarch64-edge-commit.yaml
@@ -1,6 +1,7 @@
 otk.version: "1"
 
 otk.define:
+  architecture: aarch64
   modifications:
     locale: C.UTF-8
     timezone: America/New_York
@@ -10,11 +11,11 @@ otk.define:
   packages:
     build:
       otk.external.osbuild-gen-depsolve-dnf4:
-        architecture: aarch64
+        architecture: ${architecture}
         module_platform_id: c9s
         releasever: "9"
         repositories:
-          otk.include: "common/repositories/aarch64.yaml"
+          otk.include: "common/repositories/${architecture}.yaml"
         packages:
           include:
             - coreutils
@@ -31,11 +32,11 @@ otk.define:
           exclude: []
     os:
       otk.external.osbuild-gen-depsolve-dnf4:
-        architecture: aarch64
+        architecture: ${architecture}
         module_platform_id: c9s
         releasever: "9"
         repositories:
-          otk.include: "common/repositories/aarch64.yaml"
+          otk.include: "common/repositories/${architecture}.yaml"
         packages:
           include:
             - NetworkManager
@@ -193,7 +194,7 @@ otk.target.osbuild:
               references:
                 - name:os
           options:
-            ref: centos/9/aarch64/edge
+            ref: centos/9/${architecture}/edge
             os_version: 9-stream
     - name: commit-archive
       build: name:build
--- a/example/centos/centos-9-aarch64-minimal-raw.yaml
+++ b/example/centos/centos-9-aarch64-minimal-raw.yaml
@@ -1,6 +1,7 @@
 otk.version: "1"
 
 otk.define:
+  architecture: aarch64
   modifications:
     locale: C.UTF-8
     timezone: America/New_York
@@ -10,11 +11,11 @@ otk.define:
   packages:
     build:
       otk.external.osbuild-gen-depsolve-dnf4:
-        architecture: aarch64
+        architecture: ${architecture}
         module_platform_id: c9s
         releasever: "9"
         repositories:
-          otk.include: "common/repositories/aarch64.yaml"
+          otk.include: "common/repositories/${architecture}.yaml"
         packages:
           include:
             - coreutils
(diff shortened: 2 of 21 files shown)


example/centos/centos-9-ppc64le-tar.yaml
```diff
@@ -9,7 +10,7 @@ otk.define:
   packages:
     build:
       otk.external.osbuild-gen-depsolve-dnf4:
-        architecture: s390x
```
mvo5, line 12: This was a copy/paste error before, right? I wonder why we did not spot this, I guess because the only arch specific piece in the dnf resolve is the kernel version, probably worth looking into how we can test for this in a more reliable way (i.e. tar has no kernel so we did not notice)

example/centos/centos-9-aarch64-edge-commit.yaml
```diff
@@ -1,6 +1,7 @@
 otk.version: "1"
 
 otk.define:
+  architecture: aarch64
```
mvo5, line 4: This is really interesting, the `edge-commit` between the two arches is now fully identical except for the package list

example/centos/centos-9-aarch64-minimal-raw.yaml
```diff
@@ -1,6 +1,7 @@
 otk.version: "1"
 
 otk.define:
+  architecture: aarch64
```
mvo5, line 4: And this one is also almost identical, an interesting journey

example/centos/centos-9-aarch64-tar.yaml
```diff
@@ -1,6 +1,7 @@
 otk.version: "1"
 
 otk.define:
+  architecture: aarch64
```
mvo5, line 4: AFAICT the tar image is now fully identical:
```
$ diff -u centos-9-aarch64-tar.yaml centos-9-x86_64-tar.yaml 
--- centos-9-aarch64-tar.yaml	2024-10-01 16:18:33.052687010 +0200
+++ centos-9-x86_64-tar.yaml	2024-10-01 16:18:33.052687010 +0200
@@ -1,7 +1,7 @@
 otk.version: "1"
 
 otk.define:
-  architecture: aarch64
+  architecture: x86_64
   modifications:
     locale: C.UTF-8
     timezone: America/New_York
```
I assume there will be a followup that extracts tar into a common file or something?

example/centos/centos-9-aarch64-wsl.yaml
```diff
@@ -1,14 +1,15 @@
 otk.version: "1"
 
 otk.define:
+  architecture: aarch64
```
mvo5, line 4: This is now also fully identical between the arches

example/centos/centos-9-ppc64le-tar.yaml
```diff
@@ -9,7 +10,7 @@ otk.define:
...
+        architecture: ${architecture}
         module_platform_id: c9s
         releasever: "9"
         repositories:
```
mvo5, line 16: repositories below contians still a hardcoded "ppc64le" value (but only here and not in the other arches it seems)?

mvo5 (commented):
This looks good, it's interesting how close the files become (some like tar/wsl identical now exepct for the toplevel define of the arch and some like qcow2/edge-commit extremly close).

 But there some inconsistencies left:

- fragment/grub2/x86_64-no-bios.yaml contains hardcoded x86_64 (and hardcoded kernel version)
- ppc64le hardcoded in centos-9-ppc64le-tar.yaml

Other than that this looks very nice.

example/centos/centos-9-ppc64le-tar.yaml
supakeen: Yes this was an error before! It seems like something that maybe the depsolve mocking could catch if we include (part of) the depsolve request?

example/centos/centos-9-aarch64-tar.yaml
supakeen: Yep!

example/centos/centos-9-ppc64le-tar.yaml
supakeen: Fixed. Probably misgrepped :)

supakeen:
> This looks good, it's interesting how close the files become (some like tar/wsl identical now exepct for the toplevel define of the arch and some like qcow2/edge-commit extremly close).
> 
> But there some inconsistencies left:
> 
>     * fragment/grub2/x86_64-no-bios.yaml contains hardcoded x86_64 (and hardcoded kernel version)
> 
>     * ppc64le hardcoded in centos-9-ppc64le-tar.yaml
> 
> 
> Other than that this looks very nice.

Addressed both. grub2 fragment update in separate commit.

example/centos/centos-9-ppc64le-tar.yaml
mvo5: I opened https://github.com/osbuild/images/pull/968 for this now

example/centos/centos-9-x86_64-ami.yaml
```diff
@@ -33,15 +34,15 @@ otk.define:
...
         module_platform_id: c9s
         releasever: "9"
         repositories:
-          otk.include: "common/repositories/x86_64.yaml"
+          otk.include: "common/repositories/${architecture}.yaml"
         packages:
           include:
             # https://github.com/osbuild/images/blob/v0.86.0/pkg/distro/rhel/rhel9/ami.go#L189
-            # https://github.com/osbuild/images/blob/v0.86.0/pkg/platform/x86_64.go#L37
+            # https://github.com/osbuild/images/blob/v0.86.0/pkg/platform/${architecture}.go#L37
```
mvo5, line 45: this url will also no longer work

mvo5 (approved):

PR #256, opened 2024-10-01
example/centos: x86_64 wsl use gen-curl-source

supakeen:
The x86_64 wsl omnifest was not using the gen-curl-source external for depsolving. Make it do so.


mvo5 (approved):
Thank you!

PR #257, opened 2024-10-01
example/centos: qcow2 for s390x

schuellerf:
initial version of the example for centos 9 s390x qcow2

### DIFF
--- a/example/centos/pipeline/image/s390x.yaml
+++ b/example/centos/pipeline/image/s390x.yaml
@@ -1,0 +1,44 @@
+name: image
+build: name:build
+stages:
+  otk.op.join:
+    values:
+      - otk.external.otk-make-partition-stages:
+          ${filesystem}
+      - - type: org.osbuild.copy
+          inputs:
+            root-tree:
+              type: org.osbuild.tree
+              origin: org.osbuild.pipeline
+              references:
+                - name:os
+          options:
+            paths:
+              - from: input://root-tree/
+                to: mount://-/
+          devices:
+            otk.op.join:
+              values:
+                - ${fs_options.devices}
+                # TBD this seems strange:
+                - disk:
+                    type: 'org.osbuild.loopback'
+                    options:
+                      filename: 'disk.img'
+          mounts:
+            ${fs_options.mounts}
+        - type: org.osbuild.zipl.inst
+          options:
+            kernel: ${kernel.package.version}-${kernel.package.release}.${kernel.package.arch}
+            location: ${fs_options.devices.boot.options.start}
+          devices:
+            otk.op.join:
+              values:
+                - ${fs_options.devices}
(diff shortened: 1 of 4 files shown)


schuellerf:
please take a look especially on `example/centos/pipeline/image/s390x.yaml` on `# TBD this seems strange:`

example/centos/pipeline/image/s390x.yaml
```diff
@@ -0,0 +1,59 @@
...
+                    type: 'org.osbuild.loopback'
+                    options:
+                      filename: 'disk.img'
+          mounts:
+            ${fs_options.mounts}
+        - type: org.osbuild.zipl.inst
+          options:
+            kernel: 8-2.fk1.s390x
+            location: 2048
+          devices:
```
mvo5, line 34: Please have a look at the other image pipeline examples, it should be possible to use something like
```yaml
          devices:
            ${fs_options.devices}
          mounts:
            ${fs_options.mounts}
```
here.

With all the hardcoded sizes in the fragement this will fail as soon as any of the size changes so these need to be dynamically caludlated via an external (and there is support for this already).

example/centos/pipeline/image/s390x.yaml
```diff
@@ -0,0 +1,59 @@
...
+                # TBD this seems strange:
+                - disk:
+                    type: 'org.osbuild.loopback'
+                    options:
+                      filename: 'disk.img'
+          mounts:
+            ${fs_options.mounts}
+        - type: org.osbuild.zipl.inst
+          options:
+            kernel: 8-2.fk1.s390x
```
mvo5, line 32: This should use something like `${kernel.package.version}-${kernel.package.release}.${kernel.package.arch}`, please check the grub2 fragement

mvo5 (commented):
Thanks for working on this!

example/centos/pipeline/image/s390x.yaml
schuellerf: As commented in the PR in "images" I wasn't aware that the `fs_options.*` structure is fully compatible with `org.osbuild.zipl`. Thanks for pointing this out!

schuellerf: code updated, please check

example/centos/pipeline/image/s390x.yaml
schuellerf: Thanks!
I also assumed this is variable is the correct "location":
`location: ${fs_options.devices.boot.options.start}`

schuellerf:
Update: Tests here are waiting for https://github.com/osbuild/images/pull/963 to be merged

supakeen (approved):
Looks good enough to me; misses some of the deduplication stuff that has been done while this was in flight but we can fix those in followup.

PR #260, opened 2024-10-02
src/otk/context.py: improve error on missing entry in a dictionary

schuellerf:
until now the error message was like this:

`otk.error.TransformVariableLookupError: could not find ['fs_options', 'devices', 'device', 'options', 'start']`

with this PR the error helps in finding which fragment is the problem:

`otk.error.TransformVariableLookupError: could not resolve 'fs_options.devices.device.options.start' as 'device' is not defined`

Do you think this change is useful?

### DIFF
--- a/src/otk/context.py
+++ b/src/otk/context.py
@@ -100,7 +100,7 @@ class CommonContext(Context):
         for i, part in enumerate(parts):
             if isinstance(value, dict):
                 if part not in value:
-                    raise TransformVariableLookupError(f"could not find {parts!r}")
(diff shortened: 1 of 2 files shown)


supakeen (commented):
I think it is useful but it might clash with the PR @mvo5 just opened in #263?

supakeen (approved):
Seems to me like it doesn't clash since this is only changing the text.

mvo5:
> Seems to me like it doesn't clash since this is only changing the text.

It did clash :( I will resolve

PR #262, opened 2024-10-02
container: fix build

supakeen:
Build with a container that has Go 1.22 which is required by `images`.


schuellerf (approved):
nice, thanks!
I don't know why there was a distro mixture in the example I used :-(

PR #263, opened 2024-10-02
otk: improve error handling by adding "state" to all otk  errors

mvo5:
This commit centralizes error formating into the OTKError and
adds support for takingthe parser "state". This allows us to
consistently give better error information like what file
broke and (potentially in the future) more details.

Note that this now requires that all OTKError exceptions pass a
message.

Note that there are two untested errors left, I will try to get to them


test/data/error/00-invalid.err
```diff
@@ -1 +1 @@
-00-invalid.yaml was not found
+00-invalid.yaml: file /home/mvogt/devel/osbuild/otk/test/data/error/notexist was not found
```
supakeen, line 1: I seem to have no `mvogt` user on my system (yet) ;)

supakeen (approved):
Great addition :)

PR #264, opened 2024-10-02
doc: update

supakeen:
Update the documentation to the current state of the repository (not done).


supakeen:
Moving this to ready for review. While it is very terse it now mentions all externals. Going deeper into their formats and schemas is not the right time yet as we're still on shaky foundations for those :)

mvo5 (approved):
Thanks - this looks good, I have not double checked every input line but given that the existing docs are quite outdated this is a huge improvement so let's merge it.

PR #266, opened 2024-10-02
external: Support multiple external paths

bcl:
Split the OTK_EXTERNAL_PATH on `:` like normal shell path handling in order to enable adding multiple external paths via the environment. Also adds a test for the new behavior.


supakeen (approved):
Thank you, I think we discussed wanting to have this previously as well in either #185 or #94.

PR #267, opened 2024-10-03
example/centos: unify repositories

supakeen:


supakeen:
This seems to be failing due to missing mock depsolve output; if I regenerate the reference manifests I don't get any changes.

PR #268, opened 2024-10-04
Fix makefile returncode

schuellerf:
if a command fails in "ONESHELL" mode the script continues and
`make` possibly returns `0` instead of a failure


supakeen (approved):
I'm not entirely sure if Debian still has a symlink between the different bash implementations but seems fine.

PR #270, opened 2024-10-07
Makefile: add pyexternals target

mvo5:
When running `make external` not all externals are created right now, only the go ones. This commit ensures all are available inside the `./external` dir.

Not sure this is 100% correct, the makefile grew quite a bit please double check


schuellerf (approved):
Perfect, I already came across this missing dependency too, thanks!

schuellerf:
Maybe I'll add a follow up to avoid calling `ln` if not needed, but doesn't really hurt to run it on each invocation

supakeen (approved):

PR #271, opened 2024-10-07
otk/utils: introduce HiddenAddr* objects

schuellerf:
This is not even a draft. It just sketches the idea of the Objects `HiddenAttrList` and `HiddenAttrDict`

The Objects represent a dictionary and a list
but with hidden attributes to be able to
track the source of the data from the originating
yaml file.


mvo5 (commented):
I think this is very cool, my main use-case for this would be to have better errors, we could be super precise with this, I played a bit and we could have e.g. in test/data/error/03-lonely-keys.err
```diff
-otk file contains top-level keys ['targetless'] without a target
+otk file contains top-level keys 'targetless' at test/data/error/03-lonely-keys.yaml:6 without a target
```
I looked into if we can avoid the wrappers and use a side-store with references to the id() but that idea does not work for us as we manipulate/copy/transform the resulting objects so new python objects are created all the time and the id() mapping breaks down.

I would love a broken out version of this that does not use "explain" yet but just adds the annotations as it would be a great way to improve the error reporting.

supakeen:
The general idea is good and I agree with @mvo5 that the primary usecase for this is user friendliness through good error reporting.

Since this is a pre-draft I'll just say two things about the code since it's likely to change. I'd like it more if this was called 'Annotation'; I also don't think we need the `set_attribute`, `get_attribute` it seems to me that a single object implementing the appropriate dunders should be enough to wrap any 'complex' (I think that's the YAML terminology?) type.

schuellerf:
@supakeen I thought the two classes are needed as I wanted them to really act like list and dict (aso for isinstance…)
so the code would need really minor changes only where needed.

supakeen:
Let's not get into this deeply now; it is possible to do this with a single `Annotated` class but let's first see everything it'd need to support.

schuellerf:
I will definitely refine and remove the `# DEBUG - TBD: DELETE ME` but it now contains an example error message @mvo5 please check

schuellerf:
@mvo5 @supakeen should I clean this up (remove the `--explain` but keep the `test/data/error/03-lonely-keys.err` example to get the class merged and continue in separate PRs for more error message improvements and the real `--explain` functionality?

schuellerf:
I think keeping the example error message makes sense as a starting point

schuellerf:
(btw. I can not reproduce the `mypy` and `pylint` errors locally - I'll continue investigation)

supakeen (commented):
Since this is now on non-draft state, some remarks from me before I look into the code more deeply:

- Let's maybe call `.utils` -> `.annotation`.
- Let's maybe call `HiddenAttrDict` -> `AnnotatedDict`.
- Let's maybe call `HiddenAttrList` -> `AnnotatedList`.

I'll think about consolidation and sharing of protocol between those two types in the morning.

src/otk/transform.py
```diff
@@ -53,6 +54,42 @@ def construct_mapping(self, node, deep=False):
...
+    obj.set_attribute(key, "column", column)
+
+
+def hidden_attr_dict_constructor(loader, node):
+    data = loader.construct_mapping(node, deep=True)
+    hidden_data_node = HiddenAttrDict(data)
+    for key, _ in hidden_data_node.items():
+        key_data = _find_data(node.value, key)
+        _add_hidden_attributes(hidden_data_node, key, key_data)
+    return hidden_data_node
```
supakeen, line 83: These *feel* like `@classmethod`'s for their respective `AnnotatedType`'s.

src/otk/transform.py
```diff
@@ -53,6 +54,42 @@ def construct_mapping(self, node, deep=False):
...
+
+
+def hidden_attr_list_constructor(loader, node):
+    data = loader.construct_sequence(node, deep=True)
+    hidden_data_node = HiddenAttrList(data)
+    _add_hidden_attributes(hidden_data_node, "self", node)
```
supakeen, line 89: For example, this could be `HiddenAttrList.with_hidden_attributes(data, hidden_data_node, "self", node)` or such if the above as a classmethod.

src/otk/utils.py
```diff
@@ -0,0 +1,102 @@
...
+        return data
+    if isinstance(data, dict):
+        return HiddenAttrDict({key: _deep_convert(value) for key, value in data.items()})
+    if isinstance(data, list):
+        return HiddenAttrList([_deep_convert(item) for item in data])
+    return data
+
+
+class HiddenAttrList(list):
+    def __init__(self, *args, **kwargs):
```
supakeen, line 17: Why not have `__setitem__`/`__getitem__` and the slice dunders here to mirror how stuff is done in the dict variant?

src/otk/transform.py
schuellerf: The idea was that the `HiddenAttr*` objects are generic and this is a usecase specific type/usage
That's why it's in `transform.py` but for sure `transform.py` looks nicer when this is a classmethod

src/otk/utils.py
schuellerf: Not sure if I got your question right, the answer might be that there is no multiple inheritance so things have to be done twice as I wanted it to really behave like a real `dict` also type wise

src/otk/utils.py
```diff
@@ -0,0 +1,102 @@
+import json
+from typing import TypeVar, Generic
+
+
+def _deep_convert(data):
+    """Recursively convert nested dicts and lists into HiddenAttrDict and HiddenAttrList."""
+    if isinstance(data, (HiddenAttrDict, HiddenAttrList)):
```
mvo5, line 7: This is maybe a silly question - but why do we need this? The yaml parser will build the nesting for us AFAICT (i.e. everywhere we need the HiddenAttr* clases we have them via yaml constructions) - if I (naviely) remove this helper it, nothing in the tests break (except the one test that is designed to test this). So what am I missing?

mvo5 (commented):
Thanks, this is a very interesting feature. I added some suggestins inline, I don't really have proper time so it's all a bit terse. One thing I think we should do is have some more errors converted with tests, I suspec at least:
```
def resolve_list(ctx: Context, state: State, tree: list[Any]) -> list[Any]:
    """Resolving a list means applying the resolve function to each element in
    the list."""

    log.debug("resolving list %r", tree)

    return [resolve(ctx, state, val) for val in tree]
```
will not (yet) work because here the list compreshension will mean we loose the information from "tree" which used to be a annotated list

src/otk/utils.py
```diff
@@ -0,0 +1,102 @@
...
+        raise KeyError(f"Attribute '{attr_key}' for key '{key}' not found")
+
+    def __setattr__(self, key, value):
+        self[key] = self._convert_value(value)
+
+    def __getattr__(self, item):
+        return self[item]
+
+
+class HiddenAttrDictJSONEncoder(json.JSONEncoder):
```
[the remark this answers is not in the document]
schuellerf: yes, can be removed, is used in follow up commits which are not yet PRed ;)

src/otk/utils.py
schuellerf: This was actually ment for places in the code just working with dict and not caring about the types. e.g. initializing the class with a full dict (I thought this will be needed, but I could also stash it somewhere and pull out once really needed (other than tests which I thought make sense to make the class robust)

schuellerf:
> One thing I think we should do is have some more errors converted with tests

@mvo5 I'm fully for it - but wanted to keep the PR small really just introduceing the class and one example. Changing all error messages in later PRs - but if you like it, I'll introduce the class and change all error messages in this PR

mvo5:
> > One thing I think we should do is have some more errors converted with tests
> 
> @mvo5 I'm fully for it - but wanted to keep the PR small really just introducing the class and one example. Changing all error messages in later PRs - but if you like it, I'll introduce the class and change all error messages in this PR

Sorry, I think I was wrong here, we probably just need some tests like:
```python
import os

from otk.context import CommonContext
from otk.transform import process_include
from otk.traversal import State


TEST_YAML = """\
otk.version: 1

otk.target.osbuild:
  list:
    - 1
    - 2
  dict:
   nested_dict:
    foo: bar
"""

def test_annoate(tmp_path):
    test_yaml_path = tmp_path / "test.yaml"
    test_yaml_path.write_text(TEST_YAML)
    
    ctx = CommonContext(target_requested="osbuild")
    state = State()
    tree = process_include(ctx, state, test_yaml_path)

    assert "test.yaml:1" in tree.get_attribute("otk.version", "src")
    assert "test.yaml:4" in tree["otk.target.osbuild"].get_attribute("list", "src")
    assert "test.yaml:7" in tree["otk.target.osbuild"].get_attribute("dict", "src")
    assert "test.yaml:8" in tree["otk.target.osbuild"]["dict"].get_attribute("nested_dict", "src") 
    assert "test.yaml:4" in tree["otk.target.osbuild"]["list"].get_attribute(0, "src")
```
so validate some of the more complex structure, then we don't need to touch anything in the tests (those are probably something for later anyway because the error.py:OTKError already takes an optional state, if we pass an optional "Src/Oigin" class instead we would have only minimal work and get all errors "nice" almost for free (obviously that would be followups). I hope I am making sense.

schuellerf:
>  if we pass an optional "Src/Oigin" 

I'll rework this PR but we would just change the type, no extra parameter needed.

PR #272, opened 2024-10-10
otk/ui: introduce some sort of ui

supakeen:
We've spoken about being nice to users before; @schuellerf is working on adding a lot of annotated data to our data structures, I've spent a morning on writing a tiny bit of UI.

The idea is:
- output goes to stdout
- ui and logging goes to stderr
- when stderr is not a tty (e.g. redirected away) we don't show a ui
- the ui is everything that isn't logging, logging is still there and will intersperse with the ui

Based on #277.

### DIFF
--- a/src/otk/command.py
+++ b/src/otk/command.py
@@ -8,7 +8,7 @@ from . import __version__
 from .document import Omnifest
 from . import ui
 
-log = logging.getLogger(__name__)
+log = ui.Terminal(logging.getLogger(__name__))
 
 
 def root() -> int:
@@ -29,7 +29,7 @@ def run(argv: List[str]) -> int:
         print(f"otk {__version__}")
         return 0
 
-    ui.motd()
+    log.motd()
 
     if arguments.command == "compile":
         return compile(arguments)
@@ -52,7 +52,7 @@ def _process(arguments: argparse.Namespace, dry_run: bool) -> int:
     else:
         paths = extra + [pathlib.Path(arguments.input)]
 
-    ui.print(f"Compiling {path}")
+    log.info(f"Compiling {path}")
 
     # First pass of resolving the otk file is "shallow", it will not run
     # externals and not resolve anything under otk.target.*
@@ -75,7 +75,7 @@ def _process(arguments: argparse.Namespace, dry_run: bool) -> int:
         log.fatal("requested target %r does not exist in INPUT", target_requested)
         return 1
 
-    ui.print(f"Selected the {target_requested} target")
+    log.info(f"Selected the {target_requested} target")
 
     # Now do the real resolve that takes the target into account. It needs
     # a full run so that resolving includes works correctly.
@@ -83,7 +83,7 @@ def _process(arguments: argparse.Namespace, dry_run: bool) -> int:
                                for arg in ["duplicate-definition", "all"])
     doc = Omnifest(paths, target=target_requested, warn_duplicated_defs=warn_duplicated_defs)
 
-    ui.print("Done")
+    log.info("Done")
 
     # and then output by writing to the output
     if not dry_run:
--- a/src/otk/command.py
+++ b/src/otk/command.py
@@ -52,7 +52,7 @@ def _process(arguments: argparse.Namespace, dry_run: bool) -> int:
     else:
         paths = extra + [pathlib.Path(arguments.input)]
 
-    log.info(f"Compiling {path}")
+    log.info(f"Compiling {paths}")
 
     # First pass of resolving the otk file is "shallow", it will not run
     # externals and not resolve anything under otk.target.*
--- a/src/otk/command.py
+++ b/src/otk/command.py
@@ -6,6 +6,7 @@ from typing import List
 
 from . import __version__
 from .document import Omnifest
+from . import ui
 
 log = logging.getLogger(__name__)
 
@@ -49,6 +50,8 @@ def _process(arguments: argparse.Namespace, dry_run: bool) -> int:
     else:
         paths = extra + [pathlib.Path(arguments.input)]
 
+    ui.print(f"Compiling {path}")
+
     # First pass of resolving the otk file is "shallow", it will not run
     # externals and not resolve anything under otk.target.*
     #
@@ -70,11 +73,15 @@ def _process(arguments: argparse.Namespace, dry_run: bool) -> int:
         log.fatal("requested target %r does not exist in INPUT", target_requested)
         return 1
 
+    ui.print(f"Selected the {target_requested} target")
+
     # Now do the real resolve that takes the target into account. It needs
     # a full run so that resolving includes works correctly.
     warn_duplicated_defs = any(arg in getattr(arguments, "warn", [])
                                for arg in ["duplicate-definition", "all"])
     doc = Omnifest(paths, target=target_requested, warn_duplicated_defs=warn_duplicated_defs)
+
+    ui.print("Done")
 
     # and then output by writing to the output
     if not dry_run:
--- a/src/otk/command.py
+++ b/src/otk/command.py
@@ -29,6 +29,8 @@ def run(argv: List[str]) -> int:
         print(f"otk {__version__}")
         return 0
 
+    ui.motd()
+
     if arguments.command == "compile":
         return compile(arguments)
     if arguments.command == "validate":
--- a/src/otk/ui.py
+++ b/src/otk/ui.py
@@ -5,6 +5,8 @@ import threading
 from typing import Optional, Type
 from types import TracebackType
 
+from . import __version__
+
 
 class _Spinner:
     """Render a spinner with an optional prompt to stderr, if stderr isn't a
@@ -58,3 +60,8 @@ def print(text: str) -> None:  # pylint: disable=redefined-builtin
     if sys.stderr.isatty():
         sys.stderr.write(text + "\n")
         sys.stderr.flush()
+
+
+def motd() -> None:
+    # Note, this is the local print function
+    print(f"otk {__version__} -- https://www.osbuild.org/")
--- a/src/otk/ui.py
+++ b/src/otk/ui.py
@@ -6,7 +6,7 @@ from typing import Optional, Type
 from types import TracebackType
 
 
-class Spinner:
+class _Spinner:
     """Render a spinner with an optional prompt to stderr, if stderr isn't a
     tty this only prints the prompt."""
 
@@ -47,3 +47,7 @@ class Spinner:
         self.thread.join()
 
         return True
+
+
+def spinner(prompt: str) -> _Spinner:
+    return _Spinner(prompt)
--- a/src/otk/ui.py
+++ b/src/otk/ui.py
@@ -51,3 +51,10 @@ class _Spinner:
 
 def spinner(prompt: str) -> _Spinner:
     return _Spinner(prompt)
+
+
+def print(text: str) -> None:  # pylint: disable=redefined-builtin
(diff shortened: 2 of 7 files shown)


supakeen:
Might need better locking and signal handling around the thread but I didn't encounter any issues so far. We can always move all output into a single object and have it lock.

mvo5 (commented):
Some quick high-level thoughts:
- would be nice to unify the log/ui concepts, otherwise I see a lot of "ui.print(some-info")";log.info("some-info") in the code (also verbose probably means the same for ui/logs)
- not print if it's not a tty is unusal IME, most tools seem to not do spinners or color but not printing anything is rare (I think?)
- tests :)

supakeen:
> Some quick high-level thoughts:
> 
> * would be nice to unify the log/ui concepts, otherwise I see a lot of "ui.print(some-info")";log.info("some-info") in the code (also verbose probably means the same for ui/logs)

Yes, however; see the answer for the second point.

> * not print if it's not a tty is unusal IME, most tools seem to not do spinners or color but not printing anything is rare (I think?)

Yes! So in my case the UI is meant to be not-important, only for human eyes. Normal log messages are always shown (interspersed with the UI) and any redirection turns off everything related to the UI. How do you feel?

> * tests :)

Of course! (UI's always fun to test).

src/otk/ui.py
```diff
@@ -0,0 +1,72 @@
...
+    if sys.stderr.isatty():
+        sys.stderr.write(text + "\n")
+        sys.stderr.flush()
+
+
+def motd() -> None:
+    if sys.stderr.isatty():
+        sys.stderr.write(r"""       _   _
+  ___ | |_| | __
+ / _ \| __| |/ /   version: 0.1
```
schuellerf, line 67: the version can be easily fetched from the package, right?

schuellerf:
A real "spinner" would be nice :-D 
but that's probably an overkill
```
diff --git a/src/otk/ui.py b/src/otk/ui.py
index 3fd3496..410438a 100644
--- a/src/otk/ui.py
+++ b/src/otk/ui.py
@@ -1,3 +1,4 @@
+import os
 import sys
 import time
 import threading
@@ -13,20 +14,36 @@ class _Spinner:
     active: bool
     thread: threading.Thread
 
+    spin_chars = '⠋⠙⠹⠸⢰⣰⣠⣄⣆⡆⠇⠏'
+
     def __init__(self, prompt: str) -> None:
         self.prompt = prompt
 
+    @classmethod
+    def is_ansi_terminal(cls):
+        term = os.getenv('TERM', '')
+        return term.startswith('xterm') or 'ansi' in term or 'vt100' in term
+
     def _loop(self):
-        sys.stderr.write(self.prompt)
+        sys.stderr.write(self.prompt + " ")
         sys.stderr.flush()
 
         # No spinner business if we aren't printing to tty's.
         if sys.stderr.isatty():
+            i = 0
+            ansi = _Spinner.is_ansi_terminal()
             while self.active:
-                sys.stderr.write(".")
+                if ansi:
+                    i = (i + 1) % len(_Spinner.spin_chars)
+                    sys.stderr.write(_Spinner.spin_chars[i] + '\x1b[D')
+                else:
+                    sys.stderr.write(".")
                 sys.stderr.flush()
                 time.sleep(0.1)
 
+        sys.stderr.write("\x1b[D. ")
+        sys.stderr.flush()
+
         # Only write a newline if the prompt was actually there
         if self.prompt:
             sys.stderr.write("\n")
```

src/otk/ui.py
supakeen: In the future, yes. We don't have the plumbing set up for version numbers yet.

supakeen:
There are a lot of ways to make the UI prettier, I think the core issue for this right now is my decision to have a UI that is separate from log output and which gets disabled on any redirection. If we decide that that's an OK thing to do then I'll add some tests, we can merge, and we can iterate.

schuellerf (dismissed):
I'm fine with anything a little more verbose "for humans"

schuellerf:
although some tests are missing, right? :-)

mvo5:
> > Some quick high-level thoughts:
> > 
> > * would be nice to unify the log/ui concepts, otherwise I see a lot of "ui.print(some-info")";log.info("some-info") in the code (also verbose probably means the same for ui/logs)
> 
> Yes, however; see the answer for the second point.
> 
> > * not print if it's not a tty is unusal IME, most tools seem to not do spinners or color but not printing anything is rare (I think?)
> 
> Yes! So in my case the UI is meant to be not-important, only for human eyes. Normal log messages are always shown (interspersed with the UI) and any redirection turns off everything related to the UI. How do you feel?
[..]

I think it is preferable to have a single way to do output, I'm not sure there is a thing as "not important output", either it is worth showing or it is not. Also I think building blocks that are orthogonal are important, anytime something is kinda-similar to a different concept we already have I feel uneasy. With this class we have now logging that does output and a new ui that does output. And depending on context (isatty()) some is turned on or off. And depending on verbosity level in logging some is turned on and some is turned off. 

In many ways this is very similar to logging, we want to control "debug", "warn" etc and a way to increase or decrease verbosity. 
So my suggestion would be to have a UI/Output/Logger class that implements some abstract methods like "Output.Print()", "Output.{Start,Stop}Spinner()" etc. Then on non-tty system the spinner would just do nothing. We could set verbosity level with that, we could have a multiplexer to have syslog output and tty output etc.

mvo5 (commented):
It looks like there are some commits in this PR that are not related to the title of the PR, is that intentional?

supakeen:
> It looks like there are some commits in this PR that are not related to the title of the PR, is that intentional?

Yes, it needs rebasing again as it was based on top of #277 which just got merged.

src/otk/ui.py
```diff
@@ -0,0 +1,67 @@
...
+
+        return True
+
+
+def spinner(prompt: str) -> _Spinner:
+    return _Spinner(prompt)
+
+
+def print(text: str) -> None:  # pylint: disable=redefined-builtin
+    """Write a line of text to stderr if it's attached to a tty."""
```
schuellerf, line 59: As an intermediate solution to the valuable input of @mvo5 - should we info/debug-log the text here, to avoid duplication and have those interesting messages also in the logs?
( could as well be the "else" of `isatty()`? )

supakeen:
Indeed nothing changed. I'll be taking this PR in the direction you mentioned in the comment.

supakeen:
I've adjusted this PR to have a `ui.Terminal` that wraps a `logging.Logger` instance. It a `spinner` and `motd` methods right now. Spinner does nothing if `not isatty()`.

This allows us to take control of the delegated things and act as a filter and/or formatter (colors, etc) for output based on the settings of `ui.Terminal` or `isatty()`.

src/otk/ui.py
```diff
@@ -0,0 +1,69 @@
...
+    def motd(self) -> None:
+        # Note, this is the local print function
+        self.log.info(f"otk {__version__} -- https://www.osbuild.org/")
+
+    def spinner(self, prompt: str) -> _Spinner:
+        return _Spinner(prompt)
+
+    # Everything not defined is delegated to our underlying `Logger` instance,
+    # this allows `Terminal` to be used as a `Logger` wrapper.
+    def __getattr__(self, name: str) -> Any:
```
mvo5, line 68: This gives the Terminal class now the entire API surface of a logger (whch is a lot) and it's potentially confusing because we can do stuff like "addHandler" to it because but that is only honored for some stuff (everything that happens to come from logger) but not for the spinner which just writes to stderr.

My suggestion would be to do the opposite, hide the logger from the user, provide a minimal API (debug, print, spin). Also I think tests would be really good

src/otk/ui.py
```diff
@@ -0,0 +1,69 @@
...
+        exc_val: Optional[BaseException],
+        exc_tb: Optional[TracebackType],
+    ) -> bool:
+        self.active = False
+        self.thread.join()
+
+        return True
+
+
+class Terminal:
```
mvo5, line 55: Fwiw, I am not sure this is a good name, as it is now this is a logger with two extra methods, so maybe OtkLogger or something?

src/otk/ui.py
```diff
@@ -0,0 +1,69 @@
...
+    def __init__(self, log: logging.Logger):
+        self.log = log
+
+    def motd(self) -> None:
```
mvo5, line 59: This method is mixing concerns, this is named "Terminal" which indicates its a generic class. But the "modd" is very much otk specific. it seems better to provide this from the outside but if it should be here it could be OtkLogger or OtkSpinner?

src/otk/ui.py
```diff
@@ -0,0 +1,69 @@
...
+    def __init__(self, prompt: str) -> None:
+        self.prompt = prompt
+
+    def _loop(self):
+        sys.stderr.write(self.prompt)
+        sys.stderr.flush()
+
+        # No spinner business if we aren't printing to tty's.
+        if sys.stderr.isatty():
+            while self.active:
```
mvo5, line 28: (super nitpick) I wonder if these two lines should be flipped around (and the wait needs to be there of course). As it is now the code on a non-terminal will write "running foo" and then immediately "\n". Flipping around would mean it still waits without output and only when finished writes the \n (i.e. _loop does not really loop but exit right away).

src/otk/command.py
```diff
@@ -49,6 +52,8 @@ def _process(arguments: argparse.Namespace, dry_run: bool) -> int:
     else:
         paths = extra + [pathlib.Path(arguments.input)]
 
+    log.info(f"Compiling {paths}")
```
mvo5, line 55: Fwiw, AIUI info is not shown by default (which may or may not what we want here) - having a simple smoke test would make the intend clear :)

mvo5 (commented):
One concern I have here is that if I run this with `-v` I get somewhat inconsitent output now:
```
$ OTK_EXTERNAL_PATH=./external/ PYTHONPATH=./src/ python3 -m otk -v compile ./example/centos/centos-9-x86_64-ami.yaml >/dev/null
INFO:otk.command:otk 0.0.0 -- https://www.osbuild.org/
INFO:otk.command:Compiling [PosixPath('example/centos/centos-9-x86_64-ami.yaml')]
Executing external osbuild-gen-depsolve-dnf4...........
Executing external osbuild-gen-depsolve-dnf4...
Executing external osbuild-get-dnf4-package-info.
Executing external osbuild-gen-partition-table.
Executing external osbuild-make-partition-mounts-devices.
INFO:otk.command:Selected the osbuild target
Executing external osbuild-gen-depsolve-dnf4....................
Executing external osbuild-gen-depsolve-dnf4...
Executing external osbuild-get-dnf4-package-info.
Executing external osbuild-gen-partition-table.
Executing external osbuild-make-partition-mounts-devices.
Executing external osbuild-make-depsolve-dnf4-rpm-stage.
Executing external osbuild-make-depsolve-dnf4-rpm-stage.
Executing external osbuild-make-fstab-stage.
Executing external osbuild-make-partition-stages.
Executing external osbuild-make-grub2-inst-stage.
Executing external osbuild-make-depsolve-dnf4-curl-source.
INFO:otk.command:Done
```
this feels a bit uneven (i.e. some have logging prefixes, some don't)

The key feature here seems to be the spinner, how about just adding it directly via:
```
with ui.spinner("foo"):
  doo_stuff()
```
and leave the rest of "ui" for when there is more time to write/discuss this.

(really sorry that I have so many comments, hope this is still useful)

supakeen:
I'll postpone the UI since it is not necessary at the moment; this PR can later be used when we find the right abstraction level for this to live in.

PR #273, opened 2024-10-11
container: unsplit build containers

supakeen:
After the introduction of `pyexternals` as a dependency to `external` in the Makefile targets the Go container also wants to build Python externals.

Let's instead turn everything into a single build container.


schuellerf (approved):
LGTM, still we might want to call `make container` in `test.yml`

PR #274, opened 2024-10-11
github/workflows/test: add test to build the container

schuellerf:
Avoid regressions where building the container breaks
Followup of #273 
let's see how much this slows down the tests

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -28,3 +28,15 @@ jobs:
           pip install .[dev]
       - name: Run unit tests
         run: PYTHONPATH=./src make test
+
+  test-container-build:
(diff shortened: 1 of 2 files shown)


schuellerf:
…set to draft - we really don't need to build the container twice - I'll update

schuellerf:
@supakeen please check if you
* want to add this in general
* want to spend those 4min. of CI to avoid regressions
* want to set this as required or not

supakeen:
Well, personally I don't feel it adds that much; 4 minutes of course isn't the worst. @mvo5 what do you think about building the container on each PR to catch regressions in the Makefile?

supakeen (approved):

PR #280, opened 2024-10-14
otk/utils: introduce Annotated* objects

schuellerf:
The Objects (in `src/otk/annotation.py`) represent all basic types
but with "hidden" annotations to be able to
track the source of the data from the originating
yaml file.

Some of the annotations are already used in error messages, other error messages can be extended in the future with this information.

This is replacing the initial approach #271

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -33,3 +33,4 @@ repos:
     hooks:
       - id: pylint
         additional_dependencies: ["PyYAML", "types-PyYAML", "pytest"]
+        args: ["--init-hook", "import sys; sys.path.insert(0, './src')"]
--- a/src/otk/annotation.py
+++ b/src/otk/annotation.py
@@ -1,0 +1,253 @@
+import copy
+import os
+import pathlib
+from typing import TypeVar, Generic, Any
+
+
+class AnnotatedBase:
+    """ Implements a base class for annotated objects.
+    The annotations are basically a dict[str, any] of properties to attach to the class.
+    """
+
+    # base directory to calculate relative paths (easier to read)
+    _basedir = os.path.curdir
+
+    def __init__(self) -> None:
+        self._otk_annotations: dict[str, Any] = {}
+        if isinstance(self, list):
+            for idx, value in enumerate(self):
+                self[idx] = self.deep_convert(value)
+        elif isinstance(self, dict):
+            for key, value in self.items():
+                self[key] = self.deep_convert(value)
+
+    def get_annotation(self, key: str, default: Any | None = None) -> Any | None:
+        if key in self._otk_annotations:
+            return self._otk_annotations[key]
+        return default
+
+    def set_annotation(self, key: str, val: Any) -> Any:
+        self._otk_annotations[key] = val
+
+    def get_annotations(self) -> dict[str, Any]:
+        """Returns all annotations"""
+        return self._otk_annotations
+
+    def set_annotations(self, annotations: dict[str, Any]) -> None:
+        """ set all annotations """
+        self._otk_annotations = annotations
+
+    @classmethod
+    def deep_convert(cls, data: Any) -> Any:
+        """Recursively convert nested dicts and lists into AnnotatedDict and HiddenAttrList."""
+        ret = data
+        if isinstance(data, AnnotatedBase):
+            return ret
+        if isinstance(data, dict):
+            ret = AnnotatedDict({key: cls.deep_convert(value) for key, value in data.items()})
+        elif isinstance(data, list):
+            ret = AnnotatedList([cls.deep_convert(item) for item in data])
+        elif isinstance(data, str):
+            ret = AnnotatedStr(data)
+        elif isinstance(data, bool):
+            ret = AnnotatedBool(data)
+        elif isinstance(data, int):
+            ret = AnnotatedInt(data)
+        elif isinstance(data, float):
+            ret = AnnotatedFloat(data)
+
+        # note: "complex" is not implemented as it's not used in YAML
+
+        return ret
+
+    @classmethod
+    def deep_dump(cls, data: Any) -> Any:
+        """ Converting to builtin types is usually not necessary.
+            This special conversion is mainly for json.dumps(). json library needs "bool" to pass the
+            condition "<var> is True" which does not apply to AnnotatedBool,
+            so we'll just convert all
+        """
+        ret = data
+        if isinstance(data, AnnotatedDict):
+            ret = {key: cls.deep_dump(value) for key, value in data.items()}
+        elif isinstance(data, AnnotatedList):
+            ret = [cls.deep_dump(item) for item in data]
+        elif isinstance(data, AnnotatedBool):
+            ret = data.value
+        elif isinstance(data, AnnotatedFloat):
+            ret = float(data)
+        elif isinstance(data, AnnotatedInt):
+            ret = int(data)
+        elif isinstance(data, AnnotatedStr):
+            ret = str(data)
+
+        return ret
+
+    def add_source_attributes(self, key_data, prefix=""):
+        line_number = key_data.start_mark.line + 1
+        line_number_end = key_data.end_mark.line + 1
+        column = key_data.start_mark.column + 1
+        column_end = key_data.end_mark.column + 1
+        filename = os.path.relpath(key_data.start_mark.name, self._basedir)
+        self.set_annotation(f"{prefix}src", f"{filename}:{line_number}")
+        self.set_annotation(f"{prefix}filename", filename)
+        self.set_annotation(f"{prefix}line_number", line_number)
+        self.set_annotation(f"{prefix}line_number_end", line_number_end)
+        self.set_annotation(f"{prefix}column", column)
+        self.set_annotation(f"{prefix}column_end", column_end)
+
+    def __add__(self, o):
+        ret = self.deep_convert(super().__add__(o))  # pylint: disable=E1101
+        ret.set_annotations(self.squash_annotations([self, o]))
+        return ret
+
+    def __iadd__(self, o):
+        return self.__add__(o)
+
+    def __radd__(self, o):
+        return self.deep_convert(o).__add__(self)
+
+    def __deepcopy__(self, memo):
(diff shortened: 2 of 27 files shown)


supakeen (commented):
I'm still really happy about the possibilities that having this metadata unlocks in the way of error messages and user friendliness but also still very on the fence of the current approach of pretending-to-be-builtin-objects-but-not really.

Did you consider any other approaches and if so; why does the current approach fit best?

You've looked into this a *lot* more than I have but let me say very naively, would something like this:

```python
class Annotated[T]:
    value: T
```

and then passing `Annotated` around and adjusting our code to refer to `Annotated.value` fit? If we need to we can still implement all the dunders on `Annotated` but I'd prefer to avoid that.

src/otk/document.py
```diff
@@ -49,10 +49,15 @@ def ensure(cls, deserialized_data: dict[str, Any]) -> None:
             raise ParseVersionError(f"omnifest must contain a key by the name of {NAME_VERSION!r}")
 
         # no toplevel keys without a target or an otk directive
-        targetless_keys = [key for key in deserialized_data
-                           if not key.startswith(PREFIX)]
+        targetless_keys = [
+            f" * \"{key}\" in {deserialized_data[key].get_annotation('src')}"
```
bcl, line 53: If I drop in a key, `goobers:` into the top level I get a traceback from this:

```
  File "/root/otk/src/otk/document.py", line 38, in __init__
    Omnifest.ensure(tree)
  File "/root/otk/src/otk/document.py", line 53, in ensure
    f" * \"{key}\" in {deserialized_data[key].get_annotation('src')}"
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'get_annotation'
```
so it probably also needs a None check :)

schuellerf: I ran all possible tests and examples - how did you do that?

schuellerf:
> You've looked into this a _lot_ more than I have but let me say very naively, would something like this:
> 
> ```python
> class Annotated[T]:
>     value: T
> ```
I'm pretty sure that with this approach you would have to re-implement even more, including all handling of the tree, write-to-file operations and json serialization to pass to externals etc. 🤔

supakeen:
> @supakeen
> 
> > You've looked into this a _lot_ more than I have but let me say very naively, would something like this:
> > ```python
> > class Annotated[T]:
> >     value: T
> > ```
> 
> I'm pretty sure that with this approach you would have to re-implement even more, including all handling of the tree, write-to-file operations and json serialization to pass to externals etc. 🤔

What would change concretely in the handling of the tree? The write-to-file and JSON serialization would be be a 5-line thing and adjusting callsites of `json.dump(s)` so I don't see those as major problems.
 
> In my current implementation it mainly was changing the types and some special cases like list comprehension

schuellerf:
@supakeen I did some small unit-test-like things, your approach could also work.
The code (when using the value/annotations) might be less readable but
the typesystem is for sure easier with your approach as it's not trying to be transparent

supakeen:
@schuellerf I do like explicitness. Let's wait for @mvo5 to chime in as well if he feels that that might be a better approach.

schuellerf:
@supakeen @mvo5 I put up the "other" version as a "wrapper class" as Simon suggested to try.
I like the type override (this PR) better and there are less changes while using (unless you need the annotations of course)

e.g. context.py
https://github.com/osbuild/otk/pull/280/files#diff-d3985204642fa97498214ff309912ab52685064b67bbe032f3d67950e8cbed1a
here v.s
https://github.com/osbuild/otk/pull/282/files#diff-d3985204642fa97498214ff309912ab52685064b67bbe032f3d67950e8cbed1a

src/otk/annotation.py
```diff
@@ -0,0 +1,253 @@
...
+
+        new annotation keys get appended to the resulting dict
+
+        duplicate annotations will get converted to string
+        (if you need an exception for a key please implement here)
+
+        for "src" there is an exception
+        this is expected to be in the form "filename:line_number"
+        and will be sanely unified
+        """
```
achilleas-k, line 148: This is a rather poor docstring for a method that isn't very obvious or easy to read.

The first like starting with "just" doesn't help.  "Aggregates all annotations in the list" would be a good first line for it, the "just" makes it sound like things will be simple, but the function body suggests otherwise.  But I would also add "into a dictionary keyed by <something>".

"New annotation keys get appended to the dict".  Isn't every key here new?  How can there be old annotation keys?  This looks like a pure function (it's annotated as a classmethod but it's actually static right?  `cls` isn't used).

"Duplicate annotations will get converted to a string".  Does this mean unique annotations are different types?  What does that mean for the result?

"(if you need an exception for a key please implement here)".  I would change this to:
"Exceptions for specific keys are implemented here."

and then follow it with:
```
Key exceptions:
- src: ...
```

But also, I'm not entirely clear what the standard behaviour is like so I don't really understand what the exception is excepting.

src/otk/annotation.py
```diff
@@ -176,29 +177,25 @@ def squash_annotations(cls, annotation_list: list) -> dict:
 
 
 class AnnotatedList(Generic[AnnotatedListT], AnnotatedBase, list):
-    def __init__(self, seq: list[Any] | None = None) -> None:
+    def __init__(self, seq: Optional[list[Any]] = None) -> None:
```
achilleas-k, line 180: This change (and the other type annotations in this commit) doesn't seem relevant to the commit and its message.

src/otk/annotation.py
```diff
@@ -208,6 +205,18 @@ def fspath_with_include(self) -> str:
...
+if sys.version_info >= (3, 10):
+    class AnnotatedPath(AnnotatedPathBase, pathlib.Path):
+        def __init__(self, *args: Any, **kwargs: Any) -> None:
+            pathlib.Path.__init__(self, *args, **kwargs)
+            AnnotatedPathBase.__init__(self)
+else:
+    class AnnotatedPath(AnnotatedPathBase, pathlib.PosixPath):  # pylint: disable=abstract-method
+        def __init__(self, *args: Any) -> None:  # pylint: disable=W0613
+            pathlib.PosixPath.__init__(self)
+            AnnotatedPathBase.__init__(self)
```
achilleas-k, line 217: Is there a common behaviour we can use here instead of having potentially different behaviour when switching Python versions?  Does using `PosixPath` for both cause issues?
I'm looking at this and thinking it'll be a pain to troubleshoot `AnnotatedPath` behaviour if it does slightly different things based on interpreter version.

achilleas-k (commented):
I went through this a bit quickly so I only have some surface-level comments for now.

.pre-commit-config.yaml
```diff
@@ -33,3 +33,4 @@ repos:
     hooks:
       - id: pylint
         additional_dependencies: ["PyYAML", "types-PyYAML", "pytest"]
+        args: ["--init-hook", "import sys; sys.path.insert(0, './src')"]
```
mvo5, line 36: This seems unrelated to the PR(?) and could be it's own?

mvo5 (commented):
Thank you for the PR and thank you for improving the errors and the path information.

I spend a lot of time looking at this, I have many comments but I don't have the time to write them up in a nice and proper way (sorry!).  I wrote down a bunch though and I hope some are helpful.

Overall I would like that we try to simplify it more, there is a lot going on and I see the risk that the extra complexity will make our life in the future harder when we work with transform.py. The fact that we have to subclass every basic type shows the cost for the feature is high and we run into the issues like bools or None that cannot be subclassed (and `OtkBool(True) is True` will be false therefore) and we now need to be conscious about the annotations in many places make me uneasy. Even though I spend a good deal of time I don't think I spend nearly enough. I pushed a bunch of ideas into https://github.com/osbuild/otk/compare/main...mvo5:annotations-2?expand=1 which is a variant of your code/ideas with (hopefully but this is of course subjective) reduces some of the complexity by (trying) to hide some details. But that branch does not have enough tests (I think test_annotate(tmp_path) needs the various joins/merges as well for completeness) so it's hard to say if it reasonable or not (or what is missing). I added some inline comments, some/a bunch are "addressed" (or tried to get addressed) in the branch I linked to. 

Sorry that I'm so difficult and I hope this is useful nevertheless. I understand that my comments are probably quite terse and maybe hard to understand, we can have a hangout or other sync to talk in person if you prefer that and if there are questions.

src/otk/annotation.py
```diff
@@ -0,0 +1,262 @@
...
+        """Returns all annotations"""
+        return self._otk_annotations
+
+    def set_annotations(self, annotations: dict[str, Any]) -> None:
+        """ set all annotations """
+        self._otk_annotations = annotations
+
+    @classmethod
+    def deep_convert(cls, data: Any) -> Any:
+        """Recursively convert nested dicts and lists into AnnotatedDict and HiddenAttrList."""
```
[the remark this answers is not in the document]
schuellerf: Right, I think this might be obsolete when also all tests use annotated objects when interacting with internals. I wanted to collect feedback first before changing all tests.

src/otk/annotation.py
```diff
@@ -0,0 +1,262 @@
...
+        column_end = key_data.end_mark.column + 1
+        filename = os.path.relpath(key_data.start_mark.name, self._basedir)
+        self.set_annotation(f"{prefix}src", f"{filename}:{line_number}")
+        self.set_annotation(f"{prefix}filename", filename)
+        self.set_annotation(f"{prefix}line_number", line_number)
+        self.set_annotation(f"{prefix}line_number_end", line_number_end)
+        self.set_annotation(f"{prefix}column", column)
+        self.set_annotation(f"{prefix}column_end", column_end)
+
+    def __add__(self, o):
```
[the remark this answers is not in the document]
schuellerf: I think it will also be used when implementing more otk operations?

.pre-commit-config.yaml
schuellerf: in my setup this is needed to run pytest, it could be split but I need to merge this first because otherwise it fails to commit.
I'll to a separate PR, merge and rebase :-D

schuellerf:
project will be archived…

PR #282, opened 2024-10-18
[WIP] annotation wrapper class

schuellerf:
Just an alternative implementation of #280 
I don't really like the approach to have `.value` all over the place. Let's see which approach is more suitable for a future usage
This PR is just one squashed commit until we decide on which we continue.

### DIFF
--- a/.spellcheck.yml
+++ b/.spellcheck.yml
@@ -8,7 +8,7 @@ matrix:
       camel-case: true
       mode: markdown
     sources:
-      - "**/*.md|!venv/**"
+      - "**/*.md|!venv*/**"
     dictionary:
       wordlists:
         - .spellcheck-en-custom.txt
--- a/src/otk/annotation.py
+++ b/src/otk/annotation.py
@@ -1,0 +1,249 @@
+import copy
+import os
+import pathlib
+from typing import Any, Optional, TypeVar, Generic, Union
+
+# import yaml
+# from yaml._yaml import SequenceNode, ScalarNode, MappingNode
+# from yaml.resolver import BaseResolver
+
+AnnotatedNodeT = TypeVar('AnnotatedNodeT')
+
+
+class AnnotatedNode(Generic[AnnotatedNodeT]):
+
+    # to be able to print relative paths
+    _basedirs = [os.path.curdir]
+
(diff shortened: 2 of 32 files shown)


mvo5 (commented):
Thank you for this. I have some inline comments but my high level feedback would be:

A lot of what I write below is gut feeling, I don't like making suggestions based on this without some further experimentation but in the interest of time I will. I probably miss stuff (or am flat out wrong) but I *hope* the suggestions/ideas are useful.

I am not a fan of Annotated{Src,Bool,Int,Dict,...} in general, this is true for this approach and for https://github.com/osbuild/otk/pull/280. I am worried about the complexity and how we need to overload magic methods and then we have a __getitem__ on a bool etc). 

*But* I do like the idea of AnnotatedNode (or OtkNode) and AnnoatedNote.value because it gives access to the real thing and all operations will work as expected on ".value" plus the inconvenience of having to do "node.value" is actually useful as it hints the user of the API that AnnotatedNode is not the thing to work with but just a wrapper that carries the value and extra data.

I would love to experiment with something slightly smaller:
1. Just have "AnnotatedNode" with the minimal needed interface, probably hash/eq.  And AnnotatedNode takes a value and (optionally an exiting AnnotatedNode from which it will inherit the annotation history)
2. The API of AnnoatedNode would be just `.value` and `.src` with `src` a dataclass that carries a list of filename,line for now (and the two magic methods so that Nodes can be put into a dict and compared)
3. when we need to check the type (like in transform.py:resolve() we use `isinstance(data.value, dict)` 
4. we will most likely need convenience OtkDict, OtkList or navigation the tree will be extremly painful, they just need to provide: `def __getitem__(self, item): return self.value.__getitem__` and we should construct them only in the yaml loader for list/mapping (or explicitely in transform.py)

I did a little bit of experimenting in https://github.com/osbuild/otk/compare/main...mvo5:annotations-3?expand=1 - its extremly messy and incomplete but I would love to see this explored a bit more. If that works out I think it's preferable over the subclass approvch in https://github.com/osbuild/otk/pull/280 but if we really need classes for all python types here too I think the subclass one is better (but I hope we won't)

Sorry for this slightly messy review and my rambling, hope it's helpful still

PR #283, opened 2024-10-18
otk/command: fix racecondition in tests

schuellerf:
Sometimes in my local IDE testing, the lazy close
of the dst file, leads to the test failing because it's not (yet) seeing the result.
So we'll close it, when done.

### DIFF
--- a/src/otk/command.py
+++ b/src/otk/command.py
@@ -77,6 +77,8 @@ def _process(arguments: argparse.Namespace, dry_run: bool) -> int:
     # and then output by writing to the output
     if not dry_run:
         dst.write(doc.as_target_string())
+        if arguments.output is not None:
+            dst.close()
 
     return 0
 


supakeen (commented):
Which test is failing specifically? At face value this doesn't seem to be a race condition as we don't multithread and we don't have an event loop and the next line after the line you added reduces the reference count to zero by returning from the function (which generally implies immediate closing on CPython IIRC).

schuellerf:
essentially all that immediately read the file afterwards.
It's the same process ;) just a lingering / non-flushed FD in this case
https://github.com/osbuild/otk/blob/main/test/test_example.py#L22

schuellerf:
yes, I also though about just flushing - but we `open()` in that function so essentially a "deferred close" is good here.
for the `else` (that is missing in this PR) we could definitively flush stdout - that would make sense although I didn't see a test failing on that

supakeen:
I'm fine with closing the file object at the end of the function explicitly.

However I'm really confused about this, the file object should've been closed as soon as its refcount reaches zero which is when the function returns; the test executes `compile`, which executes `_process` both of which have fully returned by the time you're reading.

Can you share the traceback?

schuellerf:
There is no real traceback - the file is just empty when reading and the json parser dies.
I can send you the error afterwards - but it's just an empty file 😅
The problem is, the test opens the file again on it's own (https://github.com/osbuild/otk/blob/main/test/test_example.py#L22) when the fist FD apparently is not yet closed or flushed

supakeen:
Maybe; let's merge this to deal with the weirdness you are seeing.

supakeen (approved):

schuellerf:
Final note: this behavior only happens when running in a debugger which seems to be changing the behavior of the garbage collector.
But still it's handy to be able to run the tests from within a debugger, so thanks for merging!

src/otk/command.py
```diff
@@ -77,6 +77,8 @@ def _process(arguments: argparse.Namespace, dry_run: bool) -> int:
     # and then output by writing to the output
     if not dry_run:
         dst.write(doc.as_target_string())
+        if arguments.output is not None:
```
mvo5, line 80: Maybe it would be preferable to use `dst.flush()` here so that sys.stdout gets the same handling as files? Presumably it's the same underlying problem?

PR #286, opened 2024-10-31
externally provided variables

supakeen:
Allow defining variables from external sources such as the command-line.

This can be used for customizations (see #242) and build systems to provide, for example, repositories.

```bash
otk compile something.yaml omnifest.yaml
```

(needs additional testcase)

### DIFF
--- a/src/otk/document.py
+++ b/src/otk/document.py
@@ -19,8 +19,9 @@ class Omnifest:
     _osbuild_ctx: OSBuildContext
     _target: str
 
-    def __init__(self, path: pathlib.Path, target: str = "", *, warn_duplicated_defs: bool = False) -> None:
+    def __init__(self, paths: list[pathlib.Path], target: str = "", *, warn_duplicated_defs: bool = False) -> None:
         self._ctx = CommonContext(target_requested=target, warn_duplicated_defs=warn_duplicated_defs)
+
         # XXX: this can be removed once we find a way to deal with unset variables
         self._ctx.define("user.modifications", {})
         self._target = target
@@ -30,7 +31,10 @@ class Omnifest:
                 raise OTKError("only target osbuild supported right now")
             self._osbuild_ctx = OSBuildContext(self._ctx)
         state = State()
-        tree = process_include(self._ctx, state, path)
+
+        for path in paths:
+            tree = process_include(self._ctx, state, path)
+
         # XXX: review this code, the idea is to find top-level keys that
         # have no targets but that of course only works if there are
         # no targets in the resolving. this means we are currently forced
--- a/src/otk/command.py
+++ b/src/otk/command.py
@@ -42,10 +42,12 @@ def _process(arguments: argparse.Namespace, dry_run: bool) -> int:
         # pylint: disable=R1732
         dst = sys.stdout if arguments.output is None else open(arguments.output, "w", encoding="utf-8")
 
+    extra = [pathlib.Path(p) for p in arguments.extra] if arguments.extra else []
+
     if arguments.input is None:
-        paths = [pathlib.Path(f"/proc/self/fd/{sys.stdin.fileno()}")]
+        paths = extra + [pathlib.Path(f"/proc/self/fd/{sys.stdin.fileno()}")]
     else:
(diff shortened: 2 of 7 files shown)


mvo5:
Thanks, please also add the testcase from https://github.com/osbuild/otk/compare/main...mvo5:add-preload?expand=1 - fwiw, this will not work right now because it seems that:
```yaml
otk.version: 1

otk.define:
  default:
    modifications:
      language: en_US.UTF-8   
  
  user:
    modifications:
      language: nl_NL.UTF-8

  modifications:
    otk.op.join:
      values:
        - ${default.modifications}
        - ${user.modifications}

otk.target.osbuild:
  lang: ${modifications.language}

```
does not work today because joins seem to not substitute vars currently (but I'm not sure yet, only just noticed it)

achilleas-k:
> There is some discussion ongoing about providing the `-d` argument. Perhaps there should be only _one_ way to do this.

IMO yes, we should only have one way to do this and it should be a file (either with `--preload` as mentioned in the linked PR, or just a sequence of positional args like you mention here).  I think the general case will have more customizations/variables than a couple and having a command line option that is tailored for smaller sets with workarounds like

```shell
otk compile -d "$(< customizations.json)" omnifest.yaml
```

is the wrong way around.  We should instead have an option for feeding files of customizations and people can, if people really want to do quick and dirty things on the command line they can do
```shell
otk compile omnifest.yaml -d - <<< '{"key":"value"}'
```

But let's be deliberate with our interface choices and incentivize the cleaner approaches.  I've seen some things with `osbuild-mpp -D ...` that were really difficult to trace and work with.

supakeen:
@mvo5, @achilleas-k I have dropped the offending commit and rewritten the PR text.

src/otk/document.py
```diff
@@ -19,8 +19,14 @@ class Omnifest:
     _osbuild_ctx: OSBuildContext
     _target: str
 
-    def __init__(self, path: pathlib.Path, target: str = "", *, warn_duplicated_defs: bool = False) -> None:
+    def __init__(self, paths: list[pathlib.Path], target: str = "", *, warn_duplicated_defs: bool = False) -> None:
         self._ctx = CommonContext(target_requested=target, warn_duplicated_defs=warn_duplicated_defs)
+
+        self._ctx = CommonContext(
```
mvo5, line 25: isn't self._ctx already defined in the same way two lines above? I guess this one can go (probably leftover?)

src/otk/document.py
```diff
@@ -30,7 +36,10 @@ def __init__(self, path: pathlib.Path, target: str = "", *, warn_duplicated_defs
                 raise OTKError("only target osbuild supported right now")
             self._osbuild_ctx = OSBuildContext(self._ctx)
         state = State()
-        tree = process_include(self._ctx, state, path)
+
+        for path in paths:
```
mvo5, line 40: AFAICT what this will do is that only the last file from the commandline will produce a tree, the others will produce vars. 

Which is what we want but it might be surprising for our users, consider:
```
$ otk compile os-pipeline.otk image-pipeline.otk
```
I think this will only produce the image-pipline(?). Maybe adding `--preload-defines` or just `--extra-defines` or something would make this behavior a bit clearer for the user. The implementation would not need to change (except the arg parsing) AFAICT.

mvo5 (commented):
Code looks good, thank you! One inline question/consideration about the user interface/principle of least surprise etc. But maybe I'm missing something here, I did not dig deep or try my example out (maybe I should, sry!)

src/otk/document.py
achilleas-k: I was unsure about this too.  I see the appeal of having a list of positional args where the last one is special, but you're right, the explicitness of having a flag for these makes it feel cleaner.

src/otk/document.py
supakeen: Yea it is. Let me fix this.

src/otk/document.py
supakeen: Done.

achilleas-k (approved):

PR #287, opened 2024-11-14
example: get fedora from upstream

supakeen:
We've created an [upstream](https://pagure.io/fedora-omnifests) repository which shows a very different layout from what we've used so far. Let's submodule it in here.

It will need to get updated every now and then, we could maybe do that through CI in the future. I've filed [an issue](https://github.com/osbuild/otk/issues/288) for that.


supakeen:
It would be fine to drop the `example/fedora` repository and have it under a README as well but I'm about even about either way. The second point is true and I'll update that.

